PR #119, opened 2025-04-29
CVS-166990 Don't show unavailble dataset formats when exporting

MarkRedeman:
## 📝 Description

Before we would show all dataset formats when exporting for instance an anomaly project. This leads to some confusion as it gives the impression to a user that they might be able to change a setting that could enable the export.

JIRA: CVS-166990

New options:
localhost_3000_callback_code=hmhwablczv7wiwvrda2bjwyck state=9ab5bc1aeff040e4994c62562634c094

## ✨ Type of Change

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/project-details/components/project-dataset/export-dataset/export-dataset-dialog.component.tsx
+++ b/web_ui/src/pages/project-details/components/project-dataset/export-dataset/export-dataset-dialog.component.tsx
@@ -108,9 +108,7 @@ export const ExportDatasetDialog = ({ triggerState, datasetName }: ExportDataset
     const { prepareExportDatasetJob } = useExportDataset(datasetName);
     const hasVideos = totalVideos > 0;
 
-    const matchFormats: ExportFormatDetails[] = AVAILABLE_FORMATS.filter(
-        ({ name }) => saveVideoFormat === SaveVideoFormat.IMAGES || isDatumaroFormat(name)
-    ).filter(({ domain }) => {
+    const availableDatasetFormats = AVAILABLE_FORMATS.filter(({ domain }) => {
         const filteredDomains = [...domain];
 
         return project.domains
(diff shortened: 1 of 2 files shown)


MarkRedeman:
Note we now show "Datumaro (Recommended)" even if there is only 1 choice. I'm thinking of only showing "(Recommended)" if there is more than 1 choice.

camiloHimura (approved):

jpggvilaca (approved):

web_ui/src/pages/project-details/components/project-dataset/export-dataset/export-dataset-dialog.component.tsx
```diff
@@ -190,14 +193,12 @@ export const ExportDatasetDialog = ({ triggerState, datasetName }: ExportDataset
                                 value={exportFormat}
                                 onChange={(format: string) => setExportFormat(format as ExportFormats)}
                             >
-                                {AVAILABLE_FORMATS.map(({ name, description }: ExportFormatDetails) => {
+                                {availableDatasetFormats.map(({ name, description }: ExportFormatDetails) => {
+                                    const isDisabled = !allowedFormats.some((format) => format.name === name);
```
dwesolow, line 197: Do we need to keep this logic for `isDisabled`? I think we wanted to render only available formats and not to disable them.

web_ui/src/pages/project-details/components/project-dataset/export-dataset/export-dataset-dialog.test.tsx
```diff
@@ -135,63 +135,72 @@ describe('ExportDatasetDialog', () => {
             await renderApp({ domains: [DOMAIN.DETECTION_ROTATED_BOUNDING_BOX] });
             expect(queryRadioOption(ExportFormats.VOC)).toBeEnabled();
             expect(queryRadioOption(ExportFormats.COCO)).toBeEnabled();
-            expect(queryRadioOption(ExportFormats.YOLO)).toBeDisabled();
+            expect(queryRadioOption(ExportFormats.YOLO)).not.toBeInTheDocument();
+            expect(queryRadioOption(ExportFormats.DATUMARO)).toBeEnabled();
```
dwesolow, line 139: If we don't want to support `isDisabled` we can replace all `toBeEnabled` with `toBeVisible` and `toBeChecked` for the format that we expect to be selected by default.

web_ui/src/pages/project-details/components/project-dataset/export-dataset/export-dataset-dialog.component.tsx
MarkRedeman: In this specific case we do: if the user chooses the native video format (exporting a datasets's videos as videos instead of having an image for every annotated frame) then we want to restrict the user's choice in only choosing Datumaro.
We want the other options to still be visible though: this indicates to the user that those options are available, but they can't use native video export.

PR #123, opened 2025-04-30
CVS-163854 - replace react virtuoso - p1

camiloHimura:
## 📝 Description

This PR replaces Virtuoso with a new visualization component for the keypoint annotation list.

changes ->
-> Replace `Virtuoso` with react-aria `Virtualizer`
-> Unselect list item after using the menu options

https://github.com/user-attachments/assets/19afb9d0-35b2-4337-92f6-96702ee71f90

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
+++ b/web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
@@ -1,9 +1,11 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-import { createContext, ReactNode, useContext, useState } from 'react';
+import { createContext, ReactNode, RefObject, useContext, useRef, useState } from 'react';
 
 import { Flex, FlexProps, Grid } from '@adobe/react-spectrum';
+import { useInteractOutside } from '@react-aria/interactions';
+import { useUnwrapDOMRef } from '@react-spectrum/utils';
 import clsx from 'clsx';
 
 import classes from './list-item-grid.module.scss';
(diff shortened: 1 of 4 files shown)


pplaskie (approved):

web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
```diff
@@ -98,16 +125,21 @@ const Labels = ({ children, ...otherProps }: FlexProps) => {
...
 
     return (
-        <Flex {...otherProps} gridArea={'list-menu'}>
-            {children}
-        </Flex>
+        <>
+            <Flex {...otherProps} gridArea={'list-menu'} ref={containerRef}>
+                {children}
+            </Flex>
+            <InteractOutside containerRef={unwrapContainerRef} onClickOutside={() => setIsHovered(false)} />
```
jpggvilaca, line 141: qq: Would `useTransition` work for this? i.e. wrapping `setIsHovered` on a useTransition. We wouldnt need `onClickOutside` and would be cleaner

web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
```diff
@@ -98,16 +125,21 @@ const Labels = ({ children, ...otherProps }: FlexProps) => {
 };
 
 const ListMenu = ({ children, ...otherProps }: FlexProps) => {
-    const { isHovered } = useListItemGridContext();
+    const containerRef = useRef<null>(null);
+    const unwrapContainerRef = useUnwrapDOMRef(containerRef);
+    const { isHovered, setIsHovered } = useListItemGridContext();
```
MarkRedeman, line 130: Could this be moved into `InteractOutside`, where we'd only pass `containerRef` to `InteractOutside`? I think this could make it so that the `Flex` component would have less reasons to rerender.

MarkRedeman (approved):

web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
```diff
@@ -19,10 +21,35 @@ interface ListItemGridProps {
...
+            if (containerRef.current?.contains(target)) {
+                return;
+            }
+            // add a timeout to ensure this executes after menu-related events are processed
+            setTimeout(onClickOutside, 0);
+        },
+    });
+    return <></>;
+};
+
```
dwesolow, line 52: Could you please elaborate a bit more about `useInteractOutside` and why do we need to use it in this case?

camiloHimura: I thought the name was self-explanatory; it does something when the user interacts outside the container, in this case, updates the `isHovered` state. https://github.com/open-edge-platform/geti/pull/123#discussion_r2068406892

web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
camiloHimura: I’m afraid that won’t work, `useTransition` delays execution, but in this case, we need InteractOutside to handle user actions outside the component. For example, when clicking a menu item, it should trigger the menu action and unselect the container. The unselecting part is exactly what `InteractOutside` is meant to handle.

web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
camiloHimura: It’s possible, but `isHovered` is still needed on line 132, so I think it’s better to centralize it here. Plus, any re-render would be minimal, just the menu trigger.

PR #143, opened 2025-05-02
Add uv locks for internal libs and fix pre-commit

AlbertvanHouten:
## 📝 Description

- Add uv lock files for internal libs to ensure tests are run with the same dependencies.
- Fix pre-commit by regenerating platform service lock files. 

## ✨ Type of Change

-  🐞 **Bug fix** – Non-breaking change which fixes an issue


maxxgx (dismissed):

A-Artemis (dismissed):

JortBergfeld (dismissed):

libs/Makefile.shared-python-libs
```diff
@@ -19,7 +19,8 @@ pre-venv:
 	fi
 
 venv: pre-venv
-	uv sync ${PIP_INSTALL_PARAMS}
+	uv lock --check
```
JortBergfeld, line 22: Why first do a check if you're going to sync anyway?

AlbertvanHouten: This is mainly there for GHA. If someone adds a new dependency but does not update the lock file accordingly, it will fail the GHA for this library. We also do this for the microservices. 

We can't just rely on sync because it means our GitHub actions will run with updated dependencies while the docker images will be built with outdated dependencies from the lock.

JortBergfeld (approved):

jcchr (approved):

PR #150, opened 2025-05-05
ITEP-28186 Filter jobs by date

pplaskie:
## 📝 Description

Add possibility to filter jobs by date by select the range date.
Jobs are filtered by the creation date of the job.
PR also include basic test to check the functionality and refactor of function to create tabs in jobs-management-panel

JIRA: ITEP-28186

https://github.com/user-attachments/assets/eff5963a-c593-4e94-b4c6-4c7a14a011c1

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -27,7 +33,7 @@ interface JobsDialogProps {
 }
 
 const DEFAULT_LIMIT = 50;
-const defaultJobsCount: JobCount = {
+const DEFAULT_JOBS_COUNT: JobCount = {
```
jpggvilaca, line 36: np: here i think we should keep it camelCase because it's not a static variable. I mean, technically the object never changes, only its contents, but still, it might be misleading

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -36,22 +42,42 @@ const defaultJobsCount: JobCount = {
...
+    const RANGE_FILTER_TOOLTIP =
+        'This component filters jobs by start date of the job. For example in you select range' +
+        ' between yesterday and today it will show jobs started yesterday or today.';
     const { organizationId, workspaceId } = useWorkspaceIdentifier();
     const { useGetJobs } = useJobs({ organizationId, workspaceId });
 
-    const [projectIdFilter, setProjectIdFilter] = useState<string | undefined>();
-    const [userIdFilter, setUserIdFilter] = useState<string | undefined>();
-    const [jobTypeFilter, setJobTypeFilter] = useState<JobType[]>([]);
+    const TODAY = today(getLocalTimeZone());
```
jpggvilaca, line 51: same as above. I dont know if we should use capital case for non-static values

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -36,22 +42,42 @@ const defaultJobsCount: JobCount = {
...
+        userId: undefined as string | undefined,
+        jobTypes: [] as JobType[],
+    });
+
+    const [range, setRange] = useState<RangeValue<DateValue>>(INITIAL_DATES);
+
     const [selectedJobState, setSelectedJobState] = useState<Key>(JobState.RUNNING);
     const [sortDirection, setSortDirection] = useState<SortDirection>(SortDirection.DESC);
 
+    // Fetch jobs data
```
jpggvilaca: np: unnecessary comment

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -36,22 +42,42 @@ const defaultJobsCount: JobCount = {
...
             jobState: selectedJobState as JobState,
-            projectId: projectIdFilter,
-            jobTypes: jobTypeFilter,
-            author: userIdFilter,
+            projectId: filters.projectId,
+            jobTypes: filters.jobTypes,
+            author: filters.userId,
             limit: DEFAULT_LIMIT,
+            startTimeFrom: range.start.toString(),
+            startTimeTo: range.end.add({ days: 1 }).toString(),
```
jpggvilaca, line 77: why add 1 day? is  the filter exclusive? should we also update line 55 then?

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -60,193 +86,111 @@ export const JobsDialog = ({ isFullScreen, onClose, setIsFullScreen }: JobsDialo
...
-                    isFetchingNextPage={isFetchingNextPage}
-                    isLoading={isLoading}
-                    jobState={JobState.FAILED}
-                    setSortDirection={setSortDirection}
-                    sortDirection={sortDirection}
-                />
-            ),
-        },
+    const tabs: TabItem[] = [
+        createTab(JobState.RUNNING, 'Running jobs', numberOfRunningJobs, 'running-jobs'),
```
jpggvilaca, line 131: <3

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const DateRangePickerSmall: FC<DateRangePickerSmall> = ({ hasManualEdition, headerContent, ...props }) => {
+    const formatter = useDateFormatter({ dateStyle: 'long' });
+    const [range, setRange] = useState<RangeValue<DateValue>>();
+    const [focusedDate, setFocusedDate] = useState<DateValue | undefined>();
+
+    const rangeText = isEmpty(range)
+        ? ''
+        : formatter.formatRange(range.start.toDate(getLocalTimeZone()), range.end.toDate(getLocalTimeZone()));
+
+    useEffect(() => {
```
jpggvilaca: would it be possible to avoid this `useEffect` by moving props.value to the initialisation of the state on line 33?

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { RangeValue } from '@react-types/shared';
+import Calendar from '@spectrum-icons/workflow/Calendar';
+import isEmpty from 'lodash/isEmpty';
+import { RangeCalendarProps } from 'react-aria-components';
+
+import { QuietActionButton } from '../quiet-button/quiet-action-button.component';
+
+interface DateRangePickerSmall extends RangeCalendarProps<DateValue> {
+    hasManualEdition?: boolean;
+    headerContent?: ReactNode;
```
jpggvilaca: np: from what i can see this could be a required field, wdyt?

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
jpggvilaca: Nice. one more reason for replacing day-js :P

jpggvilaca:
Screenshot 2025-05-06 at 11 27 29
I noticed infinite renders when i open the jobs dialog. Could it have to do with the miliseconds of the date? We should make sure we only compare days and nothing else

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
pplaskie: Yes, afair it's exclusive - I'll double check because that part of code I wrote some time ago.

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
pplaskie: Value of it it's never changed, those are default values assigned when there is no data from endpoint.

pplaskie:
> Screenshot 2025-05-06 at 11 27 29 I noticed infinite renders when i open the jobs dialog. Could it have to do with the miliseconds of the date? We should make sure we only compare days and nothing else

While checking I remembered myself that it should be because of the jobs polling.

camiloHimura:
@pplaskie Can you add a video/screenshot of the new workflow?

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
pplaskie: We have similar const names capitalized in other places in the code, should I change all of them accordingly?

pplaskie:
> @pplaskie Can you add a video/screenshot of the new workflow?

Sure!

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
pplaskie: Documentation said it's inclusive, but for me it's exclusive

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.component.tsx
pplaskie: I would like to not require this field. In this case in the heading there is an information placed but maybe there will be case when header won't be needed. But I'll update condition.

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
jpggvilaca: Agree

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.component.tsx
jpggvilaca: ok

romanowska (dismissed):

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.component.tsx
```diff
@@ -0,0 +1,93 @@
+// Copyright (C) 2022-2025 Intel Corporation
```
dwesolow: We should put this component close to `jobs-actions` because it's needed only there. If we need this component later in other places, we will move it to `@geti/ui` :).

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const MOCKED_TODAY = new CalendarDate(2025, 4, 30);
+
+    const INITIAL_DATES: RangeValue<CalendarDate> = {
+        start: MOCKED_TODAY.subtract({ months: 1 }),
+        end: MOCKED_TODAY,
+    };
+
+    const mockOnChange = jest.fn();
+
+    const renderComponent = (props = {}) =>
```
dwesolow: np. `props` can be removed, we don't use them.

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+        );
+
+    afterEach(() => {
+        userEvent.keyboard('{escape}');
+    });
+
+    it('should open the date range dialog when the button is clicked', async () => {
+        renderComponent();
+
+        expect(screen.getByRole('button', { name: 'Select date range' })).toBeInTheDocument();
```
dwesolow: np. TBH we don't need to check if the button is in the document because if it is not then `userEvent.click` in line 41 will break.

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.component.tsx
```diff
@@ -0,0 +1,93 @@
...
+    useDateFormatter,
+} from '@adobe/react-spectrum';
+import { DateValue, getLocalTimeZone } from '@internationalized/date';
+import Calendar from '@spectrum-icons/workflow/Calendar';
+import isEmpty from 'lodash/isEmpty';
+import { RangeCalendarProps } from 'react-aria-components';
+
+import { QuietActionButton } from '../quiet-button/quiet-action-button.component';
+
+interface DateRangePickerSmall extends RangeCalendarProps<DateValue> {
```
dwesolow: We allow consumer to pass `focusedValue` and `onFocusChange` which will never be used because we use internal state for this. We should do `Omit<RangeCalendarProps<DateValue>, 'focusedValue' | 'onFocusChange`>;`

web_ui/src/shared/components/header/jobs-management/jobs-actions/jobs-filtering.component.tsx
```diff
@@ -95,29 +92,18 @@ export const JobsFiltering = ({ defaultValues, onChange }: JobsFilteringProps):
...
-
-        if (!!isSelectedUserInProject) return;
-
-        setSelectedUser(undefined);
-        onChange(selectedProject, undefined, selectedJobTypes);
-    }, [isLoadingUsers, onChange, selectedJobTypes, selectedProject, selectedUser, users]);
-
-    useEffect(() => {
-        onChange(selectedProject, selectedUser, selectedJobTypes);
-    }, [selectedProject, selectedUser, selectedJobTypes, onChange]);
```
dwesolow, line 120: ❤️

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -36,22 +43,42 @@ const defaultJobsCount: JobCount = {
...
+    const RANGE_FILTER_TOOLTIP =
+        'This component filters jobs by start date. For example if you select a range' +
+        ' between yesterday and today it will show jobs started yesterday or today.';
     const { organizationId, workspaceId } = useWorkspaceIdentifier();
     const { useGetJobs } = useJobs({ organizationId, workspaceId });
 
-    const [projectIdFilter, setProjectIdFilter] = useState<string | undefined>();
-    const [userIdFilter, setUserIdFilter] = useState<string | undefined>();
-    const [jobTypeFilter, setJobTypeFilter] = useState<JobType[]>([]);
+    const TODAY = today(getLocalTimeZone());
```
dwesolow, line 51: `TODAY` is not memoized and probably recreates on each render causing `INITIAL_DATES` to recreate. WDYT of defining `INITIAL_DATES` inside `useState` leveraging the callback, i.e. `useState<RangeValue<DateValue>>(() => { define and return initial dates })`, the bonus of that approach is that we could get rid of `useMemo` that does not bring value in this case.

Edit: I see `INITIAL_DATES` and `TODAY` are used below in the code. I think we could safely remove `useMemo` unless you noticed any issues with the performance so u added that.

web_ui/src/shared/components/header/jobs-management/jobs-actions/jobs-filtering.component.tsx
```diff
@@ -13,29 +13,26 @@ import { ProjectSortingOptions } from '../../../../../core/projects/services/pro
...
+    values: FiltersType;
+    onChange: (newFilters: FiltersType) => void;
 }
 
-export const JobsFiltering = ({ defaultValues, onChange }: JobsFilteringProps): JSX.Element => {
+export const JobsFiltering = ({ values, onChange }: JobsFilteringProps): JSX.Element => {
     const { organizationId, workspaceId } = useWorkspaceIdentifier();
 
-    const { projectIdFilter, userIdFilter, jobTypeFilter } = defaultValues;
+    const [jobFilters, setJobFilters] = useState<FiltersType>(values);
```
dwesolow: Do we need this state here? I see we define state in the `jobs-dialog` and we have access to that state here. IMO we could remove the `jobFilters` from this component.

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
dwesolow: Did we double check that? Maybe we/someone should correct that in the documentation?

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -60,193 +87,101 @@ export const JobsDialog = ({ isFullScreen, onClose, setIsFullScreen }: JobsDialo
         }
     );
 
-    const allJobs = getAllJobs(data);
+    const isInitialRange = useMemo(
+        () => JSON.stringify(range) === JSON.stringify(INITIAL_DATES),
+        [range, INITIAL_DATES]
+    );
+
+    const areFiltersChanged = useMemo(
```
dwesolow: We don't need `useMemo` here, all values are primitives and it's quite fast todo comparison of booleans.

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -60,193 +87,101 @@ export const JobsDialog = ({ isFullScreen, onClose, setIsFullScreen }: JobsDialo
         }
     );
 
-    const allJobs = getAllJobs(data);
+    const isInitialRange = useMemo(
+        () => JSON.stringify(range) === JSON.stringify(INITIAL_DATES),
```
dwesolow: Is this a trick not to do `range.end == INITIAL_DATES.end` etc?

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -1,20 +1,27 @@
...
+import { RangeValue } from '@react-types/shared';
 import { keepPreviousData } from '@tanstack/react-query';
 
 import { useJobs } from '../../../../core/jobs/hooks/use-jobs.hook';
 import { NORMAL_INTERVAL } from '../../../../core/jobs/hooks/utils';
-import { JobState, JobType } from '../../../../core/jobs/jobs.const';
+import { JobState } from '../../../../core/jobs/jobs.const';
 import { JobCount } from '../../../../core/jobs/jobs.interface';
+import { CornerIndicator } from '../../../../pages/annotator/annotation/annotation-filter/corner-indicator.component';
+import { RefreshButton } from '../../../../pages/annotator/components/sidebar/dataset/refresh-button/refresh-button.component';
```
dwesolow: This button comes from the `annotator` feature. Should we consider moving the button to the `shared` or the the `@geti/ui`?

web_ui/src/shared/components/date-range-picker-small/date-range-picker-small.component.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const rangeText = isEmpty(range)
+        ? ''
+        : formatter.formatRange(range.start.toDate(getLocalTimeZone()), range.end.toDate(getLocalTimeZone()));
+
+    const handleOnChange = (attribute: 'start' | 'end', value: DateValue | null | undefined) => {
+        if (isEmpty(value) || isEmpty(range)) {
+            return;
+        }
+
+        onChange && onChange({ ...range, [attribute]: value });
```
camiloHimura:
```suggestion
        isFunction(onChange) && onChange({ ...range, [attribute]: value });
```

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
pplaskie: yes, I'm changing to isEqual from lodash

web_ui/src/shared/components/header/jobs-management/jobs-actions/jobs-filtering.component.tsx
```diff
@@ -26,12 +26,10 @@ interface JobsFilteringProps {
 export const JobsFiltering = ({ values, onChange }: JobsFilteringProps): JSX.Element => {
     const { organizationId, workspaceId } = useWorkspaceIdentifier();
 
-    const [jobFilters, setJobFilters] = useState<FiltersType>(values);
```
jpggvilaca: Without this setState the filters were not updating properly (the data on the lists). But i haven't tried manually after this commit, please do check ^^

web_ui/packages/ui/src/corner-indicator/corner-indicator.component.tsx
```diff
@@ -16,7 +16,6 @@ export const CornerIndicator = ({ isActive, children }: CornerIndicatorProps): J
                 top='size-50'
                 right='size-50'
                 width='size-50'
-                height='size-50'
```
jpggvilaca: why this removal?

web_ui/src/shared/components/header/jobs-management/jobs-actions/jobs-filtering.component.tsx
pplaskie: With this was problem with reseting the filter, now there is used value from the parent component

web_ui/packages/ui/src/corner-indicator/corner-indicator.component.tsx
pplaskie: reverted

web_ui/packages/ui/index.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
+// Copyright (C) 2022-2025 Intel Corporation
```
jpggvilaca: duplicated header

jpggvilaca (dismissed):

jpggvilaca (dismissed):

MarkRedeman (dismissed):

web_ui/src/shared/components/header/jobs-management/jobs-actions/jobs-filtering.test.tsx
```diff
@@ -197,26 +189,48 @@ describe('jobs filtering', (): void => {
...
+    ])(
+        `should trigger onChange event with selected filter `,
+        async ({ filterName, filterValue, changedValues }): Promise<void> => {
+            const users = [
+                mockedUser,
+                { ...mockedUser, id: '2', email: 'admin2@example.com', firstName: 'Administrator 2' },
+                { ...mockedUser, id: '3', email: 'admin3@example.com', firstName: 'Administrator 3' },
+            ];
+
+            await renderComponent(undefined, undefined, undefined, users);
```
dwesolow, line 223: I think it'd be cleaner to pass as object instead of list of arguments. In this case you would only do `await renderComponent({ users });`.

dwesolow (dismissed):

web_ui/src/shared/components/header/jobs-management/jobs-actions/jobs-filtering.test.tsx
pplaskie: It was here like that from 100 years. I had to update test because state is keep differently.
I'm letting myself to not changing this in this pr.

jpggvilaca (approved):

MarkRedeman (approved):

dwesolow (approved):

PR #159, opened 2025-05-06
ITEP-31884 Implement API handlers for training flow revamp

dwesolow:
## 📝 Description

ITEP-31884

This PR introduces API handlers for the new configuration endpoints that will eventually replace configurable parameters. We will have
- `project_configuration` that will include piece of information such as auto training enablement, predictions suggestions enablement etc (scoped to project parameters)
- `training_configuration` that will include parameters related to OTX and parameters that will affect the training (not related to OTX) such as dataset filtering, subset splitting.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/core/configurable-parameters/dtos/configuration.interface.ts
+++ b/web_ui/src/core/configurable-parameters/dtos/configuration.interface.ts
@@ -49,9 +49,18 @@ export interface ProjectConfigurationDTO {
     task_configs: ProjectConfigurationTaskConfigsDTO[];
 }
 
-export type TrainingConfigurationDTO = {
+interface TrainingConfigurationBaseDTO {
     dataset_preparation: Record<string, ConfigurationParameterDTO[]>;
     training: ConfigurationParameterDTO[];
     evaluation: ConfigurationParameterDTO[];
+}
+
+export interface TrainingConfigurationDTO extends TrainingConfigurationBaseDTO {
     advanced_configuration?: StaticParameterDTO[];
+}
+
+export type TrainingConfigurationUpdatePayloadDTO = {
+    [K in keyof TrainingConfigurationBaseDTO]?: TrainingConfigurationBaseDTO[K] extends ConfigurationParameterDTO[]
+        ? Pick<ConfigurationParameterDTO, 'key' | 'value'>[]
+        : Record<string, Pick<ConfigurationParameterDTO, 'key' | 'value'>[]>;
 };
--- a/web_ui/src/core/configurable-parameters/dtos/configuration.interface.ts
+++ b/web_ui/src/core/configurable-parameters/dtos/configuration.interface.ts
@@ -1,0 +1,46 @@
+// Copyright (C) 2022-2025 Intel Corporation
+// LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
+
+interface ParameterBaseDTO {
+    key: string;
+    name: string;
+    description: string;
+}
+
+interface NumberParameterDTO extends ParameterBaseDTO {
+    type: 'int' | 'float';
+    value: number;
+    min_value: number;
+    max_value: number;
+    default_value: number;
+}
+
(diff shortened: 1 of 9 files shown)


web_ui/src/core/configurable-parameters/dtos/configuration.interface.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+interface ProjectConfigurationTaskConfigsDTO {
+    task_id: string;
+    training: ProjectConfigurationTaskConfigsTrainingDTO;
+    auto_training: ConfigurationParameterDTO[];
+    predictions: ConfigurationParameterDTO[];
+}
+
+export type ProjectConfigurationUploadPayloadDTO = {
```
jpggvilaca: this seems neat, but can we simplify this a little bit? At least for me this is very hard to follow.

jpggvilaca (dismissed):
LGTM

web_ui/src/core/configurable-parameters/dtos/configuration.interface.ts
MarkRedeman: I'd prefer if we can avoid any `K in keyof` shenanigans, it makes ti too hard to read during code reviews.
I.e. in this case we can replace,
```
K in keyof Pick<
        ProjectConfigurationTaskConfigsDTO,
        'training' | 'auto_training' | 'predictions'
    >
```
with
```
'training' | 'auto_training' | 'predictions'
```

Another thing I'd prefer is if we explicitely type all key/value combinations that we know will be hardcoded. For instance if we know we will have,
```
{
    "training": {
        "constraints": [
            {
                "key": "min_images_per_label",
                "value": 10
            }
        ],
    },
    "auto_training": [
        {
            "key": "enable",
            "value": false
        }
    ]
}
```

then I think I'd prefer if we type it that way:
```
interface ProjectTaskConfiguration {
    training: {
        constraints: [
            { key: "min_images_per_label", value: number },
        ]
    },
    auto_training: [
        { key: "enable", value: boolean }
    ]
}
```

this would make it easier to make sure we always pass the correct types as well as get auto completion when wanting to update a specific configuration, i.e.
```
mutate({ training: [{ key: "min_images_per_label", value: 123 }], auto_training: [{ key: "enable", value: false }] });
```
we could setup a mutation where` training`, `auto_training` or `predictions` get auto completed, and based on that we can auto complete the available keys and their values.

dwesolow: I didn't want to make it so strict because it'd be hard to extend, i.e. each time when BE adds new parameter, we would need to update the UI types. I only saw the responses examples in the design document. I have no clue if they are complete or some parameters are missing.
AFAIR until now we haven't had any issues with the current shape of the parameters that were sent, it means that components that sends that data (that handles it on the parameter level) are responsible for that and they haven't failed, i.e. if we are about to update `auto_training`, component knows what to send.

With this type of contract we gain kind of flexibility, we have strict type checking on the level of parameter groups but parameters are not strictly typed.

camiloHimura: I agree. Reading and understanding this code takes considerable effort. Extracting reusable sections could help. For example, we could create a type alias like:

`type KeyValueParam = Pick<ConfigurationParameterDTO, 'key' | 'value'>;`
Using it here and on line 73 would enhance readability, minimize repetition, and streamline future refactoring.

dwesolow: Updated

jpggvilaca (dismissed):

jpggvilaca (approved):

pplaskie (approved):

MarkRedeman (approved):

PR #191, opened 2025-05-12
fix: Don't hide quick annotation when app is no longer hovered

dwesolow:
## 📝 Description

I believe that it's regression after updating the spectrum package that we did some time ago (we replaced `useHover` from `react-aria` with our own implementation.
The issue was that the quick annotation dialog was removed from the tree when geti app was no longer hovered.
I thought about two solutions:
- move the selected media action state to the parent and don't hide children when there is media item selected. In this approach we don't need to move state and hooks to the parent like `useVideoMediaItemQuery`, `videoMediaItemForDialog`. **(I decided to go with this approach).**
- move all the necessary state together with `VideoPlayerDialog` to the parent and pass every necessary parameter to the children.

## ✨ Type of Change

- [X] 🐞 **Bug fix** – Non-breaking change which fixes an issue
- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
+++ b/web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
@@ -1,7 +1,7 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-import { useRef, useState } from 'react';
+import { Key, useRef, useState } from 'react';
 
 import { Flex, Tooltip, TooltipTrigger } from '@adobe/react-spectrum';
 import { View } from '@react-spectrum/view';
@@ -38,6 +38,7 @@ export const GridMediaItem = ({
     shouldShowAnnotationIndicator,
 }: GridMediaItemProps): JSX.Element => {
     const triggerRef = useRef(null);
+    const [selectedMediaItemAction, setSelectedMediaItemAction] = useState<Key | undefined>(undefined);
 
     const [isHovered, setIsHovered] = useState(false);
 
@@ -61,7 +62,7 @@ export const GridMediaItem = ({
             onPointerOver={handlePointerOver}
             onPointerOut={handlePointerOut}
         >
-            {(isHovered || isSelected) && (
+            {(isHovered || isSelected || selectedMediaItemAction !== undefined) && (
                 <Flex
                     wrap
                     ref={triggerRef}
@@ -103,7 +104,11 @@ export const GridMediaItem = ({
                         height={'size-500'}
                         zIndex={10}
                     >
-                        <MediaItemActions mediaItem={mediaItem} />
+                        <MediaItemActions
+                            mediaItem={mediaItem}
+                            onSelectedMediaItemActionChange={setSelectedMediaItemAction}
+                            selectedMediaItemAction={selectedMediaItemAction}
+                        />
                     </View>
                 </Flex>
             )}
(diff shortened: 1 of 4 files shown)


web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
```diff
@@ -61,7 +62,7 @@ export const GridMediaItem = ({
             onPointerOver={handlePointerOver}
             onPointerOut={handlePointerOut}
         >
-            {(isHovered || isSelected) && (
+            {(isHovered || isSelected || selectedMediaItemAction !== undefined) && (
```
MarkRedeman, line 65: Asking to clarify, so in the dialog it triggers something that makes the `MediaItemActions` no longer hovered, which then unmounts `MediaItemActions` and therefore reset its state (plus the dialog)?

web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
```diff
@@ -38,6 +38,7 @@ export const GridMediaItem = ({
     shouldShowAnnotationIndicator,
 }: GridMediaItemProps): JSX.Element => {
     const triggerRef = useRef(null);
+    const [selectedMediaItemAction, setSelectedMediaItemAction] = useState<Key | undefined>(undefined);
```
MarkRedeman, line 41: Could we be more explicit about `Key`? Now that this state is no longer in the action component, it is a bit more difficult to understand what `Key` is expected to be.
Iirc key could indicate here that the user wants to quick annotate or delete the media item?

dwesolow: I wanted to change it but it would require to change more files which relies on `Key`. For now I decided to use `Key` but we should think about refactoring the components that use our wrapper for `MenuTrigger` (it's annoying that we need to to `key.toLocaleLowercase()` etc.

web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
dwesolow: No, when u play only with the dialog it's fine, but when u want to open a new browser tab or check email (whatever) then this part is unmounted because is no longer hovered).

jpggvilaca (commented):
media-item-actions can be slightly refactored, but i think even though this fixes the issue, the real fix would be to move the dialogcontainer to the grid item. The item should contain the dialog, not the menu. Then we dont need any state checks because the item is always present. And makes more sense to me, structurally. Thoughts?

dwesolow:
> media-item-actions can be slightly refactored, but i think even though this fixes the issue, the real fix would be to move the dialogcontainer to the grid item. The item should contain the dialog, not the menu. Then we dont need any state checks because the item is always present. And makes more sense to me, structurally. Thoughts?

I don't get it, we dialog is connected with the media actions, they live together and we need to know which action was selected to decided whether we should show the dialog or now. Could u show a pseudo code with the change u mean?

jpggvilaca:
> > media-item-actions can be slightly refactored, but i think even though this fixes the issue, the real fix would be to move the dialogcontainer to the grid item. The item should contain the dialog, not the menu. Then we dont need any state checks because the item is always present. And makes more sense to me, structurally. Thoughts?
> 
> I don't get it, we dialog is connected with the media actions, they live together and we need to know which action was selected to decided whether we should show the dialog or now. Could u show a pseudo code with the change u mean?

discussed offline. my approach lacked a full overview so this fix for now has a lower cost/benefit

jpggvilaca (approved):

MarkRedeman (approved):

PR #192, opened 2025-05-12
refactor: Leverage AbortController to remove events

dwesolow:
## 📝 Description

The main advantage of this change is that we don't need to define the function outside the listener, we can use inline functions. Additionally when there are move events, it's enough to have only one abort controller to remove those events (e.g. when we have 4 event listeners, usually we'd need to have 4 remove listeners. With abort controller it's no longer the case).

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist


web_ui/src/hooks/event-listener/event-listener.hook.ts
```diff
@@ -35,22 +35,21 @@ export function useEventListener<
...
     useEffect(() => {
+        const controller = new AbortController();
         const targetElement = determineTargetElement(element);
 
         if (targetElement === null) {
             return;
         }
 
-        const eventListener: (event: EventType[EventName]) => void = (event) => {
+        targetElement.addEventListener(eventName, (event) => {
```
jpggvilaca: did i miss something or we are never removing this event listener?

jpggvilaca (commented):
What prompted this change?

web_ui/src/hooks/event-listener/event-listener.hook.ts
dwesolow: We remove it in line 52.

dwesolow:
> What prompted this change?

Self-learning. When I see that something is worth sharing and do a change and propose is in the PR ;)

PR #195, opened 2025-05-12
refactor: Leverage AbortController to remove events

dwesolow:
## 📝 Description

The main advantage of this change is that we don't need to define the function outside the listener, we can use inline functions. Additionally when there are move events, it's enough to have only one abort controller to remove those events (e.g. when we have 4 event listeners, usually we'd need to have 4 remove listeners. With abort controller it's no longer the case).

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/hooks/event-listener/event-listener.hook.ts
+++ b/web_ui/src/hooks/event-listener/event-listener.hook.ts
@@ -35,22 +35,27 @@ export function useEventListener<
     }, [handler]);
 
     useEffect(() => {
+        const controller = new AbortController();
         const targetElement = determineTargetElement(element);
 
         if (targetElement === null) {
             return;
         }
 
-        const eventListener: (event: EventType[EventName]) => void = (event) => {
-            if (savedHandler.current !== undefined) {
-                savedHandler.current(event);
+        targetElement.addEventListener(
+            eventName,
+            (event) => {
+                if (savedHandler.current !== undefined) {
+                    savedHandler.current(event as EventType[EventName]);
+                }
+            },
+            {
+                signal: controller.signal,
             }
-        };
-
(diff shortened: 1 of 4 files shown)


web_ui/src/hooks/event-listener/event-listener.hook.ts
```diff
@@ -35,22 +35,27 @@ export function useEventListener<
...
 
-        const eventListener: (event: EventType[EventName]) => void = (event) => {
-            if (savedHandler.current !== undefined) {
-                savedHandler.current(event);
+        targetElement.addEventListener(
+            eventName,
+            (event) => {
+                if (savedHandler.current !== undefined) {
+                    savedHandler.current(event as EventType[EventName]);
+                }
```
MarkRedeman, line 50: Is it possible to keep the old type hint here, i.e. use `event: EventType[EventName]` instead of using a type assertion?

MarkRedeman (approved):

web_ui/src/hooks/event-listener/event-listener.hook.ts
dwesolow: We have misstype there, I didn't spend time to investigate how/what we could change there.
image
Previously we used `EventListenerOrEventListenerObject` as type assertion  for the whole listener.

MarkRedeman: >  Previously we used EventListenerOrEventListenerObject as type assertion for the whole listener.

Ah right I had missed that. Would be nice if we can figure out how to remove it, but I guess also not worth spending a lot of time on it.

jpggvilaca (approved):

PR #203, opened 2025-05-13
Render a react-router Link when adding href to a button

MarkRedeman:
## 📝 Description

This allows us to render buttons with `<Button href="/annotator" />`, which will look like a button, but have a link behaviour: users can middle-click or control click to open the link in a new tab or click the link to open it in the current page.

We also allow passing target and rel as these can be used for implementing download links.

Note that ideally we would simply render a `<Link to="/annotator">` however this would require us to copy over the button styles so that they work for links as well.
This likely is less maintainable as it could introduce regressions whenever we update `@adobe/react-spectrum`.

## ✨ Type of Change

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/shared/components/button/button.component.tsx
+++ b/web_ui/src/shared/components/button/button.component.tsx
@@ -3,7 +3,7 @@
 
 /* eslint-disable no-restricted-imports */
 
-import { forwardRef, Ref } from 'react';
+import { ComponentProps, forwardRef, Ref } from 'react';
 
 import {
     ActionButton as SpectrumActionButton,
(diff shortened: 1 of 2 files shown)


MarkRedeman:
One issue I've found with this approach is that currently the styling does not work for some of the other button variants, for example `variant="accent"`, this is because these are styled via our custom style rules:
https://github.com/open-edge-platform/geti/blob/main/web_ui/src/theme/geti/geti-global.module.css#L2186
Removing the `button` prefix from those rules fixes it, but I found one other issue when a link is supposed to be disabled and we don't style it properly.

I can add fixes for these two scenarios in this PR or add it to another PR when we want to use such link buttons.

jpggvilaca (approved):

camiloHimura (approved):

PR #207, opened 2025-05-13
ITEP-163854 - replace react virtuoso - p3

camiloHimura:
## 📝 Description

This PR replaces `virtuoso-grid` with the `@react-aria` `virtualized` grid. The ultimate goal is to remove Virtuoso as a third-party library from the app.

The next PR will include an update on scrolling to the selected item.

**Project media list =>**

https://github.com/user-attachments/assets/78a1ec93-a635-4d9a-8be3-6080900f537a

**Annotator media list =>**

https://github.com/user-attachments/assets/03e7d539-9592-4112-99ce-c9a21866e9c1

**Test media list =>**

https://github.com/user-attachments/assets/75a70972-dc1c-46f4-8594-f1c594ac2afd

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/annotator/components/sidebar/dataset/dataset-list-item-details.component.tsx
+++ b/web_ui/src/pages/annotator/components/sidebar/dataset/dataset-list-item-details.component.tsx
@@ -44,7 +44,7 @@ export const DatasetListItemDetails = ({
     const id = getMediaId(mediaItem);
 
     const columns: ComponentProps<typeof Grid>['columns'] = useMemo(() => {
-        const initialColumns = ['var(--width)', minmax('size-1000', '1fr')];
+        const initialColumns = ['size-800', minmax('size-1000', '1fr')];
 
         if (datasetItemMenu !== undefined) {
             initialColumns.push('max-content');
@@ -64,7 +64,7 @@ export const DatasetListItemDetails = ({
                 <Grid
                     columns={columns}
                     width={'100%'}
-                    gap={'var(--gap)'}
+                    gap={'size-100'}
                     alignItems={'center'}
                     justifyContent={'center'}
                     rows={['size-800']}
--- a/web_ui/src/pages/project-details/components/project-test/test-media-items-list.component.tsx
+++ b/web_ui/src/pages/project-details/components/project-test/test-media-items-list.component.tsx
@@ -36,18 +36,20 @@ interface TestMediaItemsListProps {
 }
 
 const viewModeSettings = {
-    [ViewModes.SMALL]: { minItemSize: 90, gap: 4, maxColumns: 6 },
+    [ViewModes.SMALL]: { minItemSize: 70, gap: 4, maxColumns: 6 },
     [ViewModes.MEDIUM]: { minItemSize: 90, gap: 4, maxColumns: 4 },
-    [ViewModes.LARGE]: { minItemSize: 90, gap: 4, maxColumns: 2 },
+    [ViewModes.LARGE]: { minItemSize: 100, gap: 4, maxColumns: 2 },
     [ViewModes.DETAILS]: { size: 85, gap: 0 },
 };
 
 const getTestMediaItemId = (item: TestMediaItem) => {
+    const annotationId = 'testResult' in item ? item.testResult.annotationId : '';
+
     if (item.type === MEDIA_TYPE.IMAGE) {
-        return item.media.identifier.imageId;
+        return `${item.media.identifier.imageId}-${annotationId}`;
     }
 
-    return item.media.identifier.videoId;
+    return `${item.media.identifier.videoId}-${annotationId}`;
 };
 
 export const TestMediaItemsList = ({
@@ -120,10 +122,7 @@ export const TestMediaItemsList = ({
                             endReached={loadNextMedia}
                             mediaItems={mediaItems}
                             viewModeSettings={viewModeSettings}
-                            idFormatter={(item) => {
-                                const annotationId = 'testResult' in item ? item.testResult.annotationId : '';
-                                return `${getTestMediaItemId(item)}-${annotationId}`;
-                            }}
+                            idFormatter={getTestMediaItemId}
                             itemContent={(mediaItem) => {
                                 const mediaImageItem = mediaItem as unknown as TestImageMediaItem;
                                 const handleSelectMediaItem = () => selectMediaItem(mediaItem);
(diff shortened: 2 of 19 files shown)


web_ui/src/pages/annotator/components/sidebar/dataset/dataset-list-item-details.component.tsx
```diff
@@ -44,7 +44,7 @@ export const DatasetListItemDetails = ({
     const id = getMediaId(mediaItem);
 
     const columns: ComponentProps<typeof Grid>['columns'] = useMemo(() => {
-        const initialColumns = ['var(--width)', minmax('size-1000', '1fr')];
+        const initialColumns = ['size-800', minmax('size-1000', '1fr')];
```
MarkRedeman, line 47: Can you clarify this change?

web_ui/src/pages/utils.ts
```diff
@@ -134,3 +136,11 @@ export const projectPointOnLine = ([startPoint, endPoint]: ProjectLine, point: P
         y: b.y * scale + startPoint.y,
     };
 };
+
+export const getMediaItemId = (item: MediaItem) => {
+    if (isImage(item)) {
+        return item.identifier.imageId;
+    }
+
+    return item.identifier.videoId;
```
MarkRedeman: What if it is a video frame (I think it's possible that we receive video frames when working with the active set or some specific filters)? I believe we already have a utility that handles the three different type of media that we support.

web_ui/src/shared/components/media-items-list/media-items-list.component.tsx
```diff
@@ -1,52 +1,84 @@
...
+    ariaLabel,
+    mediaItems,
+    viewModeSettings = VIEW_MODE_SETTINGS,
+    itemContent,
+    endReached,
+    idFormatter,
+}: MediaItemsListProps<T>): JSX.Element => {
+    const config = viewModeSettings[viewMode];
+    const isDetails = viewMode === ViewModes.DETAILS;
+    const formattedMediaItems = mediaItems?.map((item) => ({ id: idFormatter(item), ...item })) ?? [];
```
MarkRedeman: Could we change the interface of `MediaItemsList` so that instead of `T[]` it takes `Array<T & { id: string }>` (or do something like `<T extends { id: string }>`?

That way we could avoid having to add the `idFormatter` as well as avoid creating a new `formattedMediaItems` array (which could become big if we are dealing with large datasets).

camiloHimura: That's a good point, but from what I’ve seen, the current dataset doesn’t use ids, only other identifiers. So I figured centralizing it here would require fewer changes across files. If we go with the id approach, we’d need to format them externally anyway.

web_ui/src/pages/annotator/components/sidebar/dataset/dataset-list-item-details.component.tsx
MarkRedeman: Nvm I saw after reviewing the rest that the way we handle the different media sizes is now changed.

web_ui/src/shared/components/media-items-list/media-items-list.component.tsx
```diff
@@ -1,52 +1,84 @@
...
-        );
-    }
-);
+                    layout={isDetails ? 'stack' : 'grid'}
+                    aria-label={ariaLabel}
+                    items={formattedMediaItems}
+                    className={classes.container}
+                >
+                    {(item) => (
+                        <ListBoxItem textValue={item.id} style={{ height: '100%' }}>
```
MarkRedeman: Can we use a non-id for the text value? E.g. perhaps we can pass in the name of the item?

web_ui/src/shared/components/media-items-list/media-items-list.component.tsx
MarkRedeman: Hmm what about in that case only using `idFormatter` where it is needed?

The thing I'd like to avoid is copying the `mediaItems` because those might be a lot bigger than the items that are being rendered.

web_ui/src/pages/annotator/components/sidebar/dataset/dataset-list-item-details.component.tsx
camiloHimura: Yes, the new grid doesn't rely on CSS variables anymore. Instead, it uses the configuration passed to the `MediaList` component

web_ui/src/pages/utils.ts
camiloHimura: Right, it now uses the `getMediaId` utility, which handles both video and frame IDs. This code was removed in my last commit.

web_ui/src/shared/components/media-items-list/media-items-list.component.tsx
camiloHimura: My last commit removed the additional formatting, This is the new approach we're using: https://github.com/open-edge-platform/geti/pull/207/files#diff-15aa145ff96eb0b7610f843256cd08e50aecf0b4a669644f2f68a148c8f57a31R79

web_ui/src/shared/components/media-items-list/media-items-list.component.tsx
camiloHimura: Since the data isn't consistent with the `name` property, I added a `getTextValue` handler to allow for a more flexible implementation.

jpggvilaca (approved):

web_ui/src/pages/project-details/components/project-test/test-media-items-list.component.tsx
```diff
@@ -34,6 +35,23 @@ interface TestMediaItemsListProps {
...
+};
+
+const getTestMediaItemId = (item: TestMediaItem) => {
+    const annotationId = 'testResult' in item ? item.testResult.annotationId : '';
+
+    if (item.type === MEDIA_TYPE.IMAGE) {
+        return `${item.media.identifier.imageId}-${annotationId}`;
+    }
+
+    return `${item.media.identifier.videoId}-${annotationId}`;
```
MarkRedeman, line 52: I'm wondering if this should not also check for a video frame number. Though I guess the `annotationId` will make the id unique so this isn't a blocker.

MarkRedeman (approved):

PR #232, opened 2025-05-16
BDD tests for removing project in training and testing dataset

JortBergfeld:
## 📝 Description

Add tests for removing project while it is training and creating a testing dataset

JIRA: ITEP-32207, ITEP-32674

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [x] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/interactive_ai/tests/e2e/features/project_removal.feature
+++ b/interactive_ai/tests/e2e/features/project_removal.feature
@@ -27,7 +27,6 @@ Feature: project removal
       When the user tries to load the image 'cat.jpg'
       Then the request is rejected
 
-  @wip
   Scenario: deletion of a project in training
     Given an annotated project of type 'detection'
     And the user requests to train a model
--- a/interactive_ai/tests/e2e/features/project_removal.feature
+++ b/interactive_ai/tests/e2e/features/project_removal.feature
@@ -31,9 +31,11 @@ Feature: project removal
     Given an annotated project of type 'detection'
     And the user requests to train a model
     And a job of type 'train' is running
+    And the user waits for 10 seconds
     When the user tries to delete the project
     Then the request is rejected
     When the user cancels the job
+    And the user waits for 5 seconds
     And the user deletes the project
     And the user tries to load the project
     Then the request is rejected
--- a/interactive_ai/tests/e2e/features/project_removal.feature
+++ b/interactive_ai/tests/e2e/features/project_removal.feature
@@ -26,3 +26,15 @@ Feature: project removal
       Then the request is rejected
       When the user tries to load the image 'cat.jpg'
       Then the request is rejected
+
+  @wip
+  Scenario: deletion of a project in training
+    Given an annotated project of type 'detection'
+    And the user requests to train a model
+    And a job of type 'train' is running
+    When the user tries to delete the project
+    Then the request is rejected
+    When the user cancels the job
+    And the user deletes the project
+    And the user tries to load the project
+    Then the request is rejected
--- a/interactive_ai/tests/e2e/features/steps/job_monitoring.py
+++ b/interactive_ai/tests/e2e/features/steps/job_monitoring.py
@@ -29,6 +29,11 @@ JOB_POLLING_PERIOD = 3  # seconds
 logger = logging.getLogger(__name__)
 
 
+@step("the user waits for {seconds:d} seconds")
+def step_wait_seconds(seconds: int) -> None:
+    time.sleep(seconds)
+
+
 @step("a job of type '{job_type:w}' is scheduled")
 def step_then_job_scheduled(context: Context, job_type: str) -> None:
     """Asserts that the "job_type" is scheduled"""
@@ -58,7 +63,6 @@ def step_then_job_running(context: Context, job_type: str) -> None:
             job_id=context.job_id,
         )
         if context.job_info.actual_instance.state == JobState.RUNNING:
-            time.sleep(10)  # provide some time for the job to properly start
             break
         time.sleep(1)
     else:
@@ -77,7 +81,6 @@ def step_when_user_cancels_job(context: Context) -> None:
         workspace_id=context.workspace_id,
         job_id=context.job_id,
     )
-    time.sleep(5)  # provide some time for the job to stop
 
 
 @step("the job completes successfully within {job_timeout:d} minutes")
--- a/interactive_ai/tests/e2e/features/steps/job_monitoring.py
+++ b/interactive_ai/tests/e2e/features/steps/job_monitoring.py
@@ -16,7 +16,7 @@ import logging
 import time
 from typing import TYPE_CHECKING
 
-from behave import then
+from behave import step
 from behave.runner import Context
 from static_definitions import JobState, JobType
 
@@ -29,7 +29,7 @@ JOB_POLLING_PERIOD = 3  # seconds
 logger = logging.getLogger(__name__)
 
 
-@then("a job of type '{job_type:w}' is scheduled")
+@step("a job of type '{job_type:w}' is scheduled")
 def step_then_job_scheduled(context: Context, job_type: str) -> None:
     """Asserts that the "job_type" is scheduled"""
     jobs_api: JobsApi = context.jobs_api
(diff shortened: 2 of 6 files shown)


interactive_ai/tests/e2e/features/project_removal.feature
```diff
@@ -26,3 +26,14 @@ Feature: project removal
...
+  Scenario: deletion of a project in training
+    Given an annotated project of type 'detection'
+    And the user requests to train a model
+    And a job of type 'train' is running
+    When the user tries to delete the project
+    Then the request is rejected
+    When the user cancels the job
+    And the user deletes the project
+    And the user tries to load the project
+    Then the request is rejected
```
maxxgx, line 41:
```suggestion
    And the user tries to delete the project
    Then the project is deleted
    And the user tries to load the project
    Then the request is rejected
```

I thinks this scenario would be clearer

interactive_ai/tests/e2e/features/steps/job_monitoring.py
```diff
@@ -45,7 +45,42 @@ def step_then_job_scheduled(context: Context, job_type: str) -> None:
...
+            organization_id=context.organization_id,
+            workspace_id=context.workspace_id,
+            job_id=context.job_id,
+        )
+        if context.job_info.actual_instance.state == JobState.RUNNING:
+            time.sleep(10)  # provide some time for the job to properly start
+            break
+        time.sleep(1)
+    else:
+        raise RuntimeError(f"A job of type '{job_type}' is not running within 20 seconds")
```
maxxgx: Can you replace the magic numbers with constants? E.g. you can use `JOB_POLLING_PERIOD` instead of `1`. Optionally, you can also introduce a timeout period like in `step_then_job_finishes_before_timeout`

interactive_ai/tests/e2e/features/steps/job_monitoring.py
```diff
@@ -45,7 +45,42 @@ def step_then_job_scheduled(context: Context, job_type: str) -> None:
...
+@step("the user cancels the job")
+def step_when_user_cancels_job(context: Context) -> None:
+    """Cancels the job"""
+    jobs_api: JobsApi = context.jobs_api
+    jobs_api.cancel_job(
+        organization_id=context.organization_id,
+        workspace_id=context.workspace_id,
+        job_id=context.job_id,
+    )
+    time.sleep(5)  # provide some time for the job to stop
```
maxxgx: Please use a constant here as well. Why is this 5s while it was 10s for `step_then_job_running`?

interactive_ai/tests/e2e/features/steps/training.py
```diff
@@ -49,7 +49,7 @@ def _train(
     context.job_id = train_response.job_id
 
 
-@when("the user requests to train a model")
+@step("the user requests to train a model")
```
leoll2, line 52: Imo it's better to avoid generic `@step` and explicitly mark each step as either a _given_, a _when_ or a _then_, because they have different semantics that also reflect in the implementation.  
Step "the user requests to train a model" is an action performed by the user, so it should be a `@when`.  
Conversely, `@given` describes a state (namely, the state of the system at the beginning of the test): such state may be the result of user actions, whereas an action is not a state on its own therefore it doesn't fit the `@given` semantics.

I suggest to rewrite the following part of your test:
```
Given an annotated project of type 'detection'
And the user requests to train a model
And a job of type 'train' is running
...
```
as:
```
Given an annotated project of type 'detection'
When the user requests to train a model
Then a job of type 'train' is running
...
```
It doesn't affect readability too much, while it retains a clear distinction between given/when/then steps.

interactive_ai/tests/e2e/features/project_removal.feature
leoll2: We use "tries to" for steps that we expect to fail (from user perspective), and in their implementation we catch the exception (error response) so that we can make assertions on it in the subsequent Then statements. In this scenario, the project is successfully deleted, so "the user deletes the project" is correct.

interactive_ai/tests/e2e/features/steps/project_management.py
```diff
@@ -364,6 +367,14 @@ def step_when_user_loads_project(context: Context) -> None:
     )
 
 
+@when("the user tries to delete the project")
```
leoll2: nitpicky: I'd move this step to L360 to immediately follow its twin step "the user deletes the project". It's slightly easier to maintain the code if a step and its "tries to" variation are next to each other.

interactive_ai/tests/e2e/features/project_removal.feature
AlbertvanHouten: I agree with Max that this logic could be a bit more clear. I think this is mainly a bit odd because it is checking that the project is successfully deleted by trying to load it. Instead, we could check that the project delete request returns a successful response. Then we can get rid of the last two steps.

interactive_ai/tests/e2e/features/steps/media_management.py
```diff
@@ -101,9 +101,27 @@ def step_when_user_loads_image(context: Context, image_name: str) -> None:
...
+
+@then("the dataset '{dataset_name}' contains an image called '{image_name}'")
+def step_then_dataset_contains_image(context: Context, dataset_name: str, image_name: str) -> None:
+    media_api: MediaApi = context.media_api
+    context._media_info_by_name[image_name] = media_api.get_image_detail(
+        organization_id=context.organization_id,
+        workspace_id=context.workspace_id,
+        project_id=context.project_id,
+        dataset_id=context._dataset_info_by_name[dataset_name].id,
+        image_id=context._media_info_by_name[image_name].id,
```
leoll2, line 126: Here you try to access `context._media_info_by_name[image_name]` while initializing it. Does it work?

interactive_ai/tests/e2e/features/steps/job_monitoring.py
```diff
@@ -45,7 +45,42 @@ def step_then_job_scheduled(context: Context, job_type: str) -> None:
...
+    expected_job_type = JobType(job_type)
+
+    for _ in range(20):
+        context.job_info = jobs_api.get_job(
+            organization_id=context.organization_id,
+            workspace_id=context.workspace_id,
+            job_id=context.job_id,
+        )
+        if context.job_info.actual_instance.state == JobState.RUNNING:
+            time.sleep(10)  # provide some time for the job to properly start
```
leoll2: I'd avoid arbitrary waits in the steps if possible, they're difficult to control and may unnecessarily slow down other tests implemented with the same steps. In your case, where you need to take into account the delay of a job to become running, I'd rather move the sleep to an explicit step:
```
When the user requests to train a model
And the user waits 10 seconds
Then a job of type 'train' is running
```
or better
```
When the user requests to train a model
Then a job of type 'train' is scheduled
When the user waits until the job is running
And the user tries to delete the project
...
```

interactive_ai/tests/e2e/features/steps/job_monitoring.py
JortBergfeld: In this case, I believe keeping 1 here is clearer, because we want to wait a maximum of 20 seconds, so the 20 in the for loop, the 1 in the sleep and the 20 seconds in the error message are all related. Creating a constant for 1 would make it harder to understand in this case. For sleeping 10 seconds to let the job really run, I have replaced it with a step "the user waits for x seconds" to make it explicit in the test.

interactive_ai/tests/e2e/features/project_removal.feature
JortBergfeld: First of all a step always fails if the response code is not successful (2xx), so if we rely on response code, the last step of the test would be "When the user deletes the project", which is a very weird ending. Secondly, this test is explicitly testing the deletion mechanism of Geti, so we cannot rely on the response code to ensure that the project was deleted. We need some additional verification that the positive response code of the project deletion endpoint indeed means that the project is deleted. Thirdly, other tests follow a similar pattern, I do agree we can probably improve the wordings of many BDD test sentences, but this should be evaluated over all tests, which would be a separate ticket.

interactive_ai/tests/e2e/features/steps/job_monitoring.py
JortBergfeld: As mentioned above, this has been replaced by an explicit step "the user waits for x seconds". The timing is based on testing, for an already running train job to lock the project, it takes more than 5 and less than 10 seconds. For a job to get cancelled it takes less than 5.

interactive_ai/tests/e2e/features/steps/training.py
JortBergfeld: I disagree, the Given keyword indicates that the step is doing something that is a prerequisite for what we want to test, the When is a keyword indicating an action that we want to test, the Then keyword indicates a sentence that tests if the expected behavior of the When action is achieved. Other than these meanings there is no real difference between these steps from a technical point of view, they are all run in the same way and should all pass in the same way. 

In this scenario we do not test whether or not the user requests to train a model actually starts a job that trains a model, we only need a job running to lock the project for the deletion. Therefore this step should be called with the Given keyword to indicate to the reader this step is necessary to set the system in a state that we need to check the proper behavior of the deletion of projects.

One limitation of `behave` is that we cannot mark a step with two different decorators, with two different sentences, `@give("the user requested to train a model")` vs `@when("the user requests to train a model")` would be nicer.

interactive_ai/tests/e2e/features/steps/media_management.py
JortBergfeld: It is initialized when creating/uploading the image, so it is no problem.

interactive_ai/tests/e2e/features/steps/project_management.py
JortBergfeld: Agreed, moved.

AlbertvanHouten (approved):

maxxgx (approved):

PR #183, opened 2025-05-09
Enable dynamic versioning for Helm charts

mramotowski:
## 📝 Description

1. Convert Chart.yaml files into templates for build-time rendering.
2. Output rendered charts to .build/ directory within each component.
3. Template version and appVersion.

## ✨ Type of Change

Select the type of change your PR introduces:

- [ ] 🐞 **Bug fix** – Non-breaking change which fixes an issue
- [ ] 🚀 **New feature** – Non-breaking change which adds functionality
- [ ] 🔨 **Refactor** – Non-breaking change which refactors the code base
- [ ] 💥 **Breaking change** – Changes that break existing functionality
- [ ] 📚 **Documentation update**
- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Describe how the changes were tested and how reviewers can test them too:

- [ ] ✅ Tested manually
- [ ] 🤖 Run automated end-to-end tests

## ✅ Checklist

Before submitting the PR, ensure the following:

- [ ] 🔍 PR title is clear and descriptive
- [ ] 📝 For internal contributors: If applicable, include the JIRA ticket number (e.g., ITEP-123456) in the PR **title**. Do **not** include full URLs
- [ ] 💬 I have commented my code, especially in hard-to-understand areas
- [ ] 📄 I have made corresponding changes to the documentation
- [ ] ✅ I have added tests that prove my fix is effective or my feature works

PR #234, opened 2025-05-16
ITEP-24513 adding Geti deployment packages for EMF 3.1

SeanCondon:
## 📝 Description

This copies over the deployment package for Geti 2.9.0 using Edge Manageability Framework 3.1 from an internal repo in to the public.

The full README is not included as there is still some debate over the location of documentation related to this method of installing Geti. A new PR will be created for documentation.

In future a CI workflow will need to Tar these charts and push it to ghcr.io using the `oras` tool. That can come in a separate PR.

 

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

- [X] 💬 I have commented my code, especially in hard-to-understand areas
- [X] 📄 I have made corresponding changes to the documentation
- [X] ✅ I have added tests that prove my fix is effective or my feature works

### DIFF
--- a/dev_tools/edge-managability-framework/deployment-package/geti-deployment-package.yaml
+++ b/dev_tools/edge-managability-framework/deployment-package/geti-deployment-package.yaml
@@ -8,41 +8,41 @@ $schema: "https://schema.intel.com/catalog.orchestrator/0.1/schema"
 name: geti
 displayName: Intel® Geti™ Platform
 description: Intel® Geti™ Computer Vision System
-version: 2.9.0
+version: 2.10.1
 
 kind: normal
 
 applications:
   - name: geti-cm-istio-csr
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-control-plane
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-impt
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-istiod
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-istio-base
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-istio-crds
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-flyte-crds
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-gateway
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-kubelet-csr-approver
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-migration-job
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-modelmesh-serving-crd
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-opa
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-prepare
-    version: 2.9.0
+    version: 2.10.1
   - name: geti-pv-creation
-    version: 2.9.0
+    version: 2.10.1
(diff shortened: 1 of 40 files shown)


dev_tools/edge-managability-framework/deployment-package/geti-deployment-package.yaml
```diff
@@ -0,0 +1,168 @@
...
+$schema: "https://schema.intel.com/catalog.orchestrator/0.1/schema"
+
+name: geti
+displayName: Intel® Geti™ Platform
+description: Intel® Geti™ Computer Vision System
+version: 2.9.0
+
+kind: normal
+
+applications:
```
mpryahin, line 15: Thanks for putting this together! A couple of things to consider so this works with our dynamic versioning:

1. **Dynamic component versions**
    
    - Geti’s images and charts are versioned at build time (we never commit those versions to VCS).
        
    - On feature branches, we only rebuild the changed components and tag them with `v{base}-{branch}-{sha}`.
        
    - Our CI also generates an umbrella chart (similar by its nature to these manifests) that “pulls in” the exact chart versions for that build. Please see an [example](https://github.com/open-edge-platform/geti/pull/183/files#diff-644860009f123e5e364beec0dad6434bd45b130a1d32bb155caf75134d938387) 
        
2. **Registry flow**
    
    - All day-to-day builds (feature/main) push to our private ECR, not GHCR.
        
    - Releases go through an RC cycle: each RC tag triggers a full rebuild of images and charts, then we promote that exact set of images and charts to GHCR if all checks pass.
        

Because of this, the manifest you’ve added can’t pin to fixed Geti versions. You have two options:

1. **Keep it separate**
    
    - Ship this package in its own repo.
        
    - Have it declare a dependency on “Geti vX.Y.Z” (the released umbrella chart) and only publish it once that version is out.
        
2. **Fold into Geti**
    
    - Add templating to your manifest so it picks up build-time values.
        
    - Resolve and inject those versions at render time (similar to how our umbrella chart works).
    
    - This approach also glues this package and Geti CI and release cycles together.
        

Either approach works; my recommendation is to keep it in a separate repo so your release cadence isn’t tied directly to Geti’s CI (to keep it simple). That way you can release your deployment package only after we’ve promoted the matching Geti version. Let me know your thoughts.

mpryahin: Just to add a couple of extra points that we discussed with @SeanCondon to support the idea of keeping this package in a separate repo:

It creates a coupling between Geti and EMF. We’d need to release Geti just to support changes on the EMF side, even if nothing changed in Geti. This would add unnecessary coordination overhead and blur the separation between the two products.

The current manifest statically references versions in GHCR, which doesn’t align with how we build and test Geti. You won’t be able to test integration with dev or feature branches unless the manifests are templated and rendered at build time with dynamic values which adds CI complexity on our side.

Keeping it standalone gives you full control of the release cycle and avoids pulling Geti into EMF timelines. Let me know what you think.

SeanCondon:
Deployment package has been moved to [App Orch Catalog Tutorials section](https://github.com/open-edge-platform/app-orch-catalog/tree/main/app-orch-tutorials) in [PR 145](https://github.com/open-edge-platform/app-orch-catalog/pull/145 )

PR #239, opened 2025-05-19
ITEP-67565: Allow users to sort labels in any order they want

JortBergfeld:
## 📝 Description

Remove label sorting from label schema, so that users can sort their labels and label groups in an order they prefer.

JIRA: [ITEP-67565](https://jira.devtools.intel.com/browse/ITEP-67565)

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/interactive_ai/libs/iai_core_py/iai_core/entities/label.py
+++ b/interactive_ai/libs/iai_core_py/iai_core/entities/label.py
@@ -145,9 +145,8 @@ class Label(PersistentEntity):
     def __eq__(self, other: object) -> bool:
         """Returns True if the two labels are equal."""
         if isinstance(other, Label):
-            return (
-                self.id_ == other.id_
-                and self.name == other.name
+            return self.id_ == other.id_ or (
+                self.name == other.name
                 and self.color == other.color
                 and self.hotkey == other.hotkey
                 and self.domain == other.domain
--- a/interactive_ai/libs/iai_core_py/iai_core/entities/label.py
+++ b/interactive_ai/libs/iai_core_py/iai_core/entities/label.py
@@ -145,13 +145,7 @@ class Label(PersistentEntity):
     def __eq__(self, other: object) -> bool:
         """Returns True if the two labels are equal."""
         if isinstance(other, Label):
-            return self.id_ == other.id_ or (
-                self.name == other.name
-                and self.color == other.color
-                and self.hotkey == other.hotkey
-                and self.domain == other.domain
-                and self.is_anomalous == other.is_anomalous
-            )
+            return self.id_ == other.id_
         return False
 
     def __lt__(self, other: object):
--- a/interactive_ai/libs/iai_core_py/iai_core/entities/label_schema.py
+++ b/interactive_ai/libs/iai_core_py/iai_core/entities/label_schema.py
@@ -13,7 +13,6 @@ from bson import ObjectId
 
 from iai_core.entities.graph import MultiDiGraph
 from iai_core.entities.label import Label
-from iai_core.entities.scored_label import ScoredLabel
 from iai_core.utils.uid_generator import generate_uid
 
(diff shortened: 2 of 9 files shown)


maxxgx:
@JortBergfeld Changing the label order in the label schema does not affect inference because we reconstruct the label ordering based on the ModelAPI labels order. See [IPredictionToAnnotationConverter](https://github.com/open-edge-platform/geti/blob/f4120fb7dfcdf56cc9bd1fef528d9ad498f53f67/interactive_ai/workflows/geti_domain/common/jobs_common_extras/evaluation/services/prediction_to_annotation_converter.py#L43)

interactive_ai/libs/iai_core_py/iai_core/entities/label.py
```diff
@@ -145,9 +145,8 @@ def __repr__(self):
     def __eq__(self, other: object) -> bool:
         """Returns True if the two labels are equal."""
         if isinstance(other, Label):
-            return (
-                self.id_ == other.id_
-                and self.name == other.name
+            return self.id_ == other.id_ or (
+                self.name == other.name
```
leoll2: This equality comparison has a weird semantics. In general, the equality should be either _entity-based_ or _value-based_. The first case is when you want to check if two objects represent the same entity: assuming that an entity is uniquely identified by its id, then the comparison should be simply on `.id_`. Example: each person is uniquely identified by a tax code; even if that person changes name or hair color, you can claim that it's still the same person. If instead you want that two objects are equivalent by _value_, then you should consider the attributes that define the value of that object. Geometric example: Degrees(10) == Degrees (370) != Degrees(45).

Labels are entities, so I think we should opt for entity-based comparison here: two Label objects are the same if the id matches. It should be safe, I don't think the rest of the code relies on mutable properties like the color.

interactive_ai/libs/iai_core_py/iai_core/entities/label_schema.py
```diff
@@ -103,7 +97,7 @@ def __eq__(self, other: object):
         """Returns True if the LabelGroup is equal to the other object."""
         if not isinstance(other, LabelGroup):
             return False
-        return self.id_ == other.id_ and (set(self.labels) == set(other.labels) and self.group_type == other.group_type)
+        return self.id_ == other.id_ or (set(self.labels) == set(other.labels) and self.group_type == other.group_type)
```
leoll2: Is this code used anywhere? I mean, do we compare label groups anywhere?

interactive_ai/libs/iai_core_py/iai_core/entities/label_schema.py
```diff
@@ -594,9 +557,8 @@ def __repr__(self) -> str:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, LabelSchema):
             return False
-        return (
-            self.id_ == other.id_
-            and self.project_id == other.project_id
+        return self.id_ == other.id_ or (
+            self.project_id == other.project_id
```
leoll2: Same as the other comment, I would reduce this comparison to ids only. For other equivalence comparisons between label schemas we have specialized methods, e.g. `_compare_labels_sync_status`

interactive_ai/libs/iai_core_py/iai_core/entities/label_schema.py
```diff
@@ -737,32 +699,10 @@ def __repr__(self) -> str:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, LabelSchemaView):
             return False
-        return (
+        return self.id_ == other.id_ or (
```
leoll2: Same as the other comment

leoll2 (approved):

maxxgx (approved):

PR #255, opened 2025-05-20
ITEP-31984 ITEP-31805 - Add GET training_configuration and PATCH project_configuration endpoints

maxxgx:
## 📝 Description

- Add `GET /training_configuration` endpoint.
- Updated `ConfigurableParametersRESTView` to support hyperparameters.
- Updated configuration tools pydantic models: 
  - Add partial models to allow for optional fields. This is very useful for PATCH endpoints when updating part of the configuration. Full model can still be used for validation.
  - Made the base pydantic models stricter with `None` type used only when the field is not supported by a model manifest.
- Add `PATCH /project_configuration` endpoint.

Tickets: ITEP-31984 ITEP-31805

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist


interactive_ai/services/director/app/communication/views/training_configuration_rest_views.py
```diff
@@ -0,0 +1,72 @@
...
+
+        :param training_configuration: training configuration
+        :param exclude_none: whether to exclude optional parameters
+        :return: REST view of the training configuration
+        """
+        if isinstance(training_configuration, NullTrainingConfiguration):
+            return {}
+
+        rest_view = {
+            "task_id": str(training_configuration.model_extra["task_id"]),
```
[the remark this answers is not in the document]
maxxgx: Note: `training_configuration.task_id` does not work

interactive_ai/services/director/app/communication/controllers/training_configuration_controller.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @classmethod
+    @unified_tracing
+    def get_configuration(
+        cls,
+        project_identifier: ProjectIdentifier,
+        task_id: ID,
+        model_manifest_id: str | None = None,
+        model_id: ID | None = None,  # noqa: ARG003
+        exclude_none: bool = False,
```
leoll2: Why is it configurable and why does it default to `False`? Based on my understanding, the `None` value is reserved to the fields that are not applicable to the model, e.g. tiling parameters for a classification model. Then the training configuration should only return params which make sense for the model.

libs/configuration_tools/src/geti_configuration_tools/training_configuration.py
```diff
@@ -58,8 +59,8 @@ class GlobalDatasetPreparationParameters(BaseModel):
     Controls data splitting and filtering before being passed for the training.
     """
 
-    subset_split: SubsetSplit = Field(title="Subset split", description="Configuration for splitting data into subsets")
-    filtering: Filtering = Field(title="Filtering", description="Configuration for filtering annotations")
+    subset_split: SubsetSplit | None = Field(default=None, title="Subset split", description="Configuration for splitting data into subsets")
+    filtering: Filtering | None = Field(default=None, title="Filtering", description="Configuration for filtering annotations")
```
leoll2: Why are these nullable?

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/hyperparameters.py
```diff
@@ -13,19 +13,26 @@ class DatasetPreparationParameters(BaseModel):
...
 
-    max_epochs: int = Field(gt=0, title="Maximum epochs", description="Maximum number of training epochs to run")
+    max_epochs: int | None = Field(
+        gt=0, default=None, title="Maximum epochs", description="Maximum number of training epochs to run"
+    )
     early_stopping: EarlyStopping | None = Field(
         default=None, title="Early stopping", description="Configuration for early stopping mechanism"
     )
-    learning_rate: float = Field(gt=0, lt=1, title="Learning rate", description="Base learning rate for the optimizer")
+    learning_rate: float = Field(gt=0.0, lt=1.0, default=0.001, title="Learning rate", description="Base learning rate for the optimizer")
```
leoll2: Why do these have a default? We expect the model manifest to always provide these, else we should throw a validation error

interactive_ai/services/director/app/communication/controllers/training_configuration_controller.py
maxxgx: The issue is that we use these Pydantic models for both task-level and algorithm-level (per model architecture) configurations, and they need to be flexible to allow for incomplete configurations. For instance, in task-level config, we might only specify values for subset_split. If we don't make the fields optional with None, we would have to provide the entire configuration, making it incompatible with the PATCH endpoint.

maxxgx: I've removed the `exclude_none` option. However, when an object field is None, it will be removed, as we use this to indicate that it's not supported. Making fields optional also allows for more flexible configuration, as mentioned earlier.

To further illustrate which None parameters are removed:
```
dataset_preparation:
    subset_split:
        ...
    filtering:
        ....
    augmentation:
        random_affine:
            enable: true
            degrees: 0
            translate_x: 0.0
            translate_y: 0.0
            scale: 1.0         
        tiling: null   <-- excluded in rest view
```

Edit: only nested object can be None (they will not be returned in the REST view). Basic types (int, float, str, bool) cannot be None and always have a default value.

libs/configuration_tools/src/geti_configuration_tools/training_configuration.py
maxxgx: As mentioned above, this is to allow for flexibility for task-level and algorithm-level configuration. Otherwise, you would always need to provide the full configuration, which would make task-level configuration pointless, as algorithm-level configurations would overwrite all of them.

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/hyperparameters.py
maxxgx: We can create a unit test in the `supported_model` package to ensure that all model manifests provide these parameters. However, we can't throw a validation error for this, as we also need to accommodate incomplete training configurations.

interactive_ai/services/director/app/communication/controllers/training_configuration_controller.py
leoll2: After offline discussion, we agreed to *never* return `null`-valued parameters in the REST response. If any parameter is `None` in the Python object, meaning that such parameter is not applicable, it should be filtered out when generating the REST response.

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/hyperparameters.py
maxxgx: I'm looking at [this solution](https://stackoverflow.com/a/76560886) to generate an optional model to use for the PATCH endpoint. If this works, we don't need make the original model fields optional anymore and therefore we can use the schema for validation

maxxgx: @leoll2 Added implementation of the partial models, which will be used for the PATCH endpoints and training configuration overlaying: `ModelManifest <-- task level configuration <-- model architecture level configuration`.  `TrainingConfigurationRepo` will persist `PartialTrainingConfiguration` entities since we rely on overlaying and  never save a full configuration object.

interactive_ai/services/director/app/communication/exceptions.py
```diff
@@ -563,3 +563,34 @@ def __init__(self, message: str) -> None:
...
+    """Exception raised when the project configuration could not be found in the database"""
+
+    def __init__(self, project_id: ID, task_id: ID | None = None) -> None:
+        message = (
+            f"The requested project configuration could not be found. Project ID: `{project_id}`, task ID: `{task_id}`"
+            if task_id
+            else ""
+        )
+        super().__init__(
+            http_status=status.HTTP_404_NOT_FOUND,
```
leoll2:
```suggestion
            http_status=http.HTTPStatus.NOT_FOUND,
```

interactive_ai/services/director/app/communication/endpoints/project_configuration_endpoints.py
```diff
@@ -38,3 +39,21 @@ def get_project_configuration(
...
+    project_configuration: PartialProjectConfiguration,
+) -> None:
+    """Update the configuration for a specific project."""
+    if not FeatureFlagProvider.is_enabled(FeatureFlag.FEATURE_FLAG_NEW_CONFIGURABLE_PARAMETERS):
+        raise GetiBaseException(
+            message="Feature not available",
+            error_code="feature_not_available",
+            http_status=HTTPStatus.FORBIDDEN,
+        )
+    ProjectConfigurationRESTController().update_configuration(
```
leoll2:
```suggestion
    ProjectConfigurationRESTController.update_configuration(
```

interactive_ai/services/director/app/communication/endpoints/project_configuration_endpoints.py
```diff
@@ -38,3 +39,21 @@ def get_project_configuration(
             http_status=HTTPStatus.FORBIDDEN,
         )
     return ProjectConfigurationRESTController().get_configuration(project_identifier=project_identifier)
+
+
+@project_configuration_router.patch("/project_configuration", status_code=HTTPStatus.NO_CONTENT)
+def update_project_configuration(
+    project_identifier: Annotated[ProjectIdentifier, Depends(get_project_identifier)],
+    project_configuration: PartialProjectConfiguration,
```
leoll2: What error does the endpoint raise if the body contains unrecognized parameters, i.e. parameters that are not present in the ProjectConfiguration model?

interactive_ai/services/director/app/communication/controllers/utils.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    for key, value in list(d.items()):
+        if isinstance(value, dict):
+            delete_none_from_dict(value)
+        elif value is None:
+            del d[key]
+        elif isinstance(value, list):
+            for v_i in value:
+                if isinstance(v_i, dict):
+                    delete_none_from_dict(v_i)
```
leoll2, line 20: In theory there's also the case where `v_i` itself is a list, but I guess it doesn't happen in practice.

interactive_ai/services/director/app/communication/controllers/utils.py
leoll2, line 1: Please add unit tests for the functions in this file

libs/configuration_tools/tests/test_project_configuration.py
```diff
@@ -121,3 +121,16 @@ def test_project_configuration(self, project_config_dict, expected_config):
             assert task_config.auto_training.min_images_per_label == (
                 expected_task_config.auto_training.min_images_per_label
             )
+
+    def test_partial_task_config(self):
```
leoll2:
```suggestion
    def test_partial_task_config(self) -> None:
```

interactive_ai/services/director/app/communication/endpoints/training_configuration_endpoints.py
```diff
@@ -0,0 +1,48 @@
...
+    dependencies=[Depends(setup_session_fastapi)],
+)
+
+
+@training_configuration_router.get("/training_configuration")
+def get_training_configuration(
+    project_identifier: Annotated[ProjectIdentifier, Depends(get_project_identifier)],
+    task_id: Annotated[str | None, Query()] = None,
+    model_manifest_id: Annotated[str | None, Query()] = None,
+    model_id: Annotated[str | None, Query()] = None,
```
leoll2, line 34: I think you can use `get_optional_model_id` from fastapi-tools here.

maxxgx: Seems not possible, `model_id` is passed as query parameter

interactive_ai/services/director/app/communication/controllers/utils.py
maxxgx: Added

interactive_ai/services/director/app/communication/endpoints/project_configuration_endpoints.py
maxxgx: It will raise an error 400 due to the pydantic model validation (extra params is not allowed).

interactive_ai/services/director/app/communication/controllers/utils.py
maxxgx: Yeah there is no nested list of list in the configuration models

interactive_ai/services/director/app/communication/controllers/training_configuration_controller.py
```diff
@@ -0,0 +1,53 @@
...
+        :param project_identifier: Identifier for the project (containing organization_id, workspace_id, and project_id)
+        :param task_id: ID of the task to retrieve configuration for
+        :param model_manifest_id: Optional ID of the model manifest to retrieve specific configurations
+        :return: Dictionary representation of the training configuration
+        :raises TaskNotFoundException: If the task does not exist
+        """
+        if not TaskNodeRepo(project_identifier).exists(task_id):
+            raise TaskNodeNotFoundException(task_node_id=task_id)
+
+        # TODO: if model_id is provided, load configuration from the model entity
```
leoll2, line 42: Is this TODO for a future PR?

leoll2 (approved):

interactive_ai/services/director/app/communication/controllers/training_configuration_controller.py
maxxgx: Yes, I'll address this in the next PR

PR #259, opened 2025-05-21
ITEP - 66906 - Replace QuietActionButton with ActionButton

jpggvilaca:
## 📝 Description

* Next ones will be AcceptButton and QuietToggleButton

REVIEW BY COMMIT

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/packages/ui/src/button/button.component.tsx
```diff
@@ -44,5 +64,8 @@ export const Button = forwardRef((props: ButtonProps, ref: ButtonProps['ref']) =
 });
 
 export const ActionButton = forwardRef((props: ActionButtonProps, ref: ActionButtonProps['ref']) => {
-    return <SpectrumActionButton {...props} ref={ref} />;
+    const { colorMode = ColorMode.DARK, UNSAFE_className, ...rest } = props;
+    const className = getActionButtonClass(colorMode, UNSAFE_className);
+
+    return <SpectrumActionButton {...rest} ref={ref} UNSAFE_className={className} />;
```
camiloHimura: This approach ensures that `getActionButtonClass` has a single responsibility and also removes the need for the nested ternary condition

```suggestion
    return <SpectrumActionButton {...rest} ref={ref} UNSAFE_className={clsx(UNSAFE_className, getActionButtonClass(colorMode))} />;
```

camiloHimura (dismissed):

web_ui/packages/ui/src/button/button.component.tsx
jpggvilaca: I like it. Updated!

web_ui/packages/ui/src/button/button.component.tsx
```diff
@@ -32,8 +35,23 @@ function LinkBuilder({ href, target, rel }: { href: string; target?: LinkProps['
 
 export interface ActionButtonProps extends SpectrumActionButtonProps {
     ref?: Ref<FocusableRefValue<HTMLElement, HTMLButtonElement>>;
+    colorMode?: ColorMode;
+}
+
+export enum ColorMode {
```
dwesolow: WDYT of doing `type ColorMode = 'dark' | 'light'`? My main motivation is that then consumer would not need to import `ColorMode` to use a dark or light version of that button (like `variant` prop in the `Button`, maybe we could call it `variant` instead of `ColorMode`?).

jpggvilaca: We could, but there are 3 variants, and we would also lose auto-completion. Perhaps after i refactor all these 3 buttons (quiet,toggle,accept) we can see if we're actually using all variants. If we don't then maybe we can make this change. You had me at "let's delete the enum"

dwesolow: Number of variants it's not a problem, we can do `type ActionButtonVariant = 'dark' | 'light' | 'blue';`.
We don't love auto-completion, IDE will show you want are available options.
At the moment (I know it's been like this for a while but I've just realised it) we use that `ColorMode` also in the `ThreeDotsFlashing` component which means `button` and `three-dots-flashing` are coupled. To decouple it, `ThreeDotsFlashing` should provide its own contract for the available variants.

jpggvilaca: Makes sense! Updated

web_ui/packages/ui/src/button/button.component.tsx
```diff
@@ -32,8 +35,19 @@ function LinkBuilder({ href, target, rel }: { href: string; target?: LinkProps['
 
 export interface ActionButtonProps extends SpectrumActionButtonProps {
     ref?: Ref<FocusableRefValue<HTMLElement, HTMLButtonElement>>;
+    colorVariant?: ButtonColorVariant;
 }
 
+type ButtonColorVariant = 'dark' | 'light' | 'blue';
```
dwesolow, line 41: np. `ActionButtonColorVariant` (not to mix with `Button`s variants)

web_ui/packages/ui/src/button/button.component.tsx
MarkRedeman: > We could, but there are 3 variants, and we would also lose auto-completion. Perhaps after i refactor all these 3 buttons (quiet,toggle,accept) we can see if we're actually using all variants. If we don't then maybe we can make this change. You had me at "let's delete the enum"

Wouldn't auto completion be even better? Now the IDE only needs to autocomplete the string versus having to first import an enum, inserting enum and then auto completing its value.

Personally I'm tend to prefer type unions a lot more than enums because of this.

web_ui/packages/ui/src/button/button.component.tsx
```diff
@@ -32,8 +35,19 @@ function LinkBuilder({ href, target, rel }: { href: string; target?: LinkProps['
...
+
+const getActionButtonClass = (colorVariant: ButtonColorVariant = 'dark') => {
+    return colorVariant === 'dark'
```
dwesolow, line 44: One idea to improve it:
```
const COLOR_VARIANTS: Record<ButtonColorVariant, string> = {
   dark: buttonClasses.actionButtonDark,
   light: buttonClasses.actionButtonLight,
   blue: buttonClasses.actionButtonBlue
};

return COLOR_VARIANTS[colorVariant];
```

web_ui/packages/ui/src/button/button.component.tsx
```diff
@@ -32,8 +35,19 @@ function LinkBuilder({ href, target, rel }: { href: string; target?: LinkProps['
 
 export interface ActionButtonProps extends SpectrumActionButtonProps {
     ref?: Ref<FocusableRefValue<HTMLElement, HTMLButtonElement>>;
+    colorVariant?: ButtonColorVariant;
```
dwesolow, line 38: WDYT of `variant`, without the `color`prefix? Like we have in `Button` component.

dwesolow (approved):
Minor comments but generally well done! :)

web_ui/packages/ui/src/button/button.component.tsx
jpggvilaca: It would be confusing, and there's also places where we render Button or ActionButton, and passing it as props would be ambiguous. I'd rather make it explicitly distinct

pplaskie (approved):

web_ui/packages/ui/src/button/button.component.tsx
jpggvilaca: Updated on my next PR

web_ui/packages/ui/src/button/button.component.tsx
jpggvilaca: Updated on my next PR

PR #249, opened 2025-05-20
:chores: Moved dataset import/export to its own feature folder

jpggvilaca:
## 📝 Description

Before:
`pages/landing-page/landing-page-workspace/components/dataset-import-to-new-project`
`shared/components/dataset-import-deletion-dialog`
`project-details/components/project-dataset/dataset-imports`
...other scattered folders

After:
Screenshot 2025-05-23 at 16 25 33

## ✨ Type of Change

Select the type of change your PR introduces:

- [ ] 🐞 **Bug fix** – Non-breaking change which fixes an issue
- [ ] 🚀 **New feature** – Non-breaking change which adds functionality
- [x] 🔨 **Refactor** – Non-breaking change which refactors the code base
- [ ] 💥 **Breaking change** – Changes that break existing functionality
- [ ] 📚 **Documentation update**
- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Describe how the changes were tested and how reviewers can test them too:

- [x] ✅ Tested manually
- [ ] 🤖 Run automated end-to-end tests

## ✅ Checklist

Before submitting the PR, ensure the following:

- [x] 🔍 PR title is clear and descriptive
- [ ] 📝 For internal contributors: If applicable, include the JIRA ticket number (e.g., ITEP-123456) in the PR **title**. Do **not** include full URLs
- [ ] 💬 I have commented my code, especially in hard-to-understand areas
- [ ] 📄 I have made corresponding changes to the documentation
- [ ] ✅ I have added tests that prove my fix is effective or my feature works

PR #289, opened 2025-05-23
ITEP-66302 Move `src/core/services` and `src/core/requests` to `@geti/core` @geti/core [PART 4]

dwesolow:
## 📝 Description

Move `src/core/services` and `src/core/requests` to `@geti/core`.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist


web_ui/packages/core/eslint.config.js
```diff
@@ -25,4 +25,20 @@ export default [
...
     },
+    {
+        files: ['./src/**/*.{ts,tsx}'],
+        rules: {
+            'no-restricted-imports': [
+                'error',
+                {
+                    patterns: [
+                        {
+                            group: ['@geti/core'],
```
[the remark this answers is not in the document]
dwesolow: Don't need this, once we move everything to `core` then we will do barrel export.

web_ui/packages/core/src/services/use-deployment-config-query.hook.ts
```diff
@@ -5,7 +5,7 @@ import { useSuspenseQuery, UseSuspenseQueryOptions } from '@tanstack/react-query
 import axios from 'axios';
 import { object, string } from 'yup';
 
-import { oidcConfigCidaas, oidcConfigIntelSSO } from './../../core/auth/configuration';
+import { oidcConfigCidaas, oidcConfigIntelSSO } from '../../../../src/core/auth/configuration';
```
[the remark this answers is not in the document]
dwesolow: We want to use relative imports when importing from the same package.

web_ui/packages/core/src/requests/query-keys.ts
```diff
@@ -1,26 +1,29 @@
 // Copyright (C) 2022-2025 Intel Corporation
```
MarkRedeman, line 1: I think we shouldn't move the query keys to a separate package. My main motivation for that is that this would then require us to update the `@geti/core` package when we want to make application level changes. I'm not sure if that's worth it.

dwesolow: Hmm, good point but then we might end up having cyclic deps, i.e. geti application that owns query keys uses `@geti/core` hooks that uses that query keys (for tanstack query). That's why I felt having query keys close to hooks/services would be better.

The other idea is to leave only services in the `core` and move api hooks to `features/api` but I'm afraid that it'd introduce cyclic deps as well.

jpggvilaca: Imho i agree with this change. Although what @MarkRedeman said is true i think this is the best option we have so far. Ideally perhaps we would have the query-keys contained in the feature itself, not in any global file (feature/some-feature/api-keys.ts). Since we cannot do this now, having it on core for now is , in my view, the best trade-off we can get right now. Once we start filling the `features` folders, we can start moving the contents of this file one piece at a time, with the end goal being to get rid of it. P.S. i could already start doing this on https://github.com/open-edge-platform/geti/pull/249 if we all agree

pplaskie (approved):

romanowska (approved):

PR #258, opened 2025-05-21
ITEP-66302 Prepare `core` a bit to move it to `@geti/core` [PART 1]

dwesolow:
## 📝 Description

1. Remove function from canvas utils that was used only in the tests, not in the app's code.
2. Remove `sort`, `sortAscending`, `sortDescending` and use `orderBy`.
3. Move configurable parameters types to core from `shared`.
4. Move dataset import local storage hook into features from `core`.
5. Combine `SortDirection` to have only one.

JIRA: ITEP-66302

## ✨ Type of Change

Select the type of change your PR introduces:

- [ ] 🐞 **Bug fix** – Non-breaking change which fixes an issue
- [ ] 🚀 **New feature** – Non-breaking change which adds functionality
- [X] 🔨 **Refactor** – Non-breaking change which refactors the code base
- [ ] 💥 **Breaking change** – Changes that break existing functionality
- [ ] 📚 **Documentation update**
- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Describe how the changes were tested and how reviewers can test them too:

- [ ] ✅ Tested manually
- [ ] 🤖 Run automated end-to-end tests

## ✅ Checklist

Before submitting the PR, ensure the following:

- [X] 🔍 PR title is clear and descriptive
- [ ] 📝 For internal contributors: If applicable, include the JIRA ticket number (e.g., ITEP-123456) in the PR **title**. Do **not** include full URLs
- [ ] 💬 I have commented my code, especially in hard-to-understand areas
- [ ] 📄 I have made corresponding changes to the documentation
- [ ] ✅ I have added tests that prove my fix is effective or my feature works

PR #297, opened 2025-05-25
gh-295 Fix not begin able to load test results

### LINKED ISSUE #295: Minified React error at Test Tab
**### Description:**
After training my model and running the test‐set evaluation with geti on Ubuntu 24.04 (via WSL2), I can’t view the “Completed Tests” page. The evaluation itself completes successfully, but as soon as I click to open the test results I get:

Error: Minified React error #31; visit
https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7Bsort_direction%7D

All other commands (training, evaluation, navigating to other pages) work normally.

Image
Image

MarkRedeman:
## 📝 Description

I think this was a regression from #258, most likely by combining a duplicated [SortDirection](https://github.com/open-edge-platform/geti/pull/258/commits/1e1c3c41b44a240bffa5160b16efccb42cae31e8#diff-c94851f3462c06404434404947f588fa5f86349e390d55f472ec246547ba96b2L15) we accidentally changed the value used when sorting test results to be `ASC` and `DESC` instead of the lower case variants.

While we had e2e tests that check the test results these tests only checked the test score and would finish before triggering the error reported in #295.

Fixes #295 

## ✨ Type of Change

- [x] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/core/media/media-filter.interface.ts
+++ b/web_ui/src/core/media/media-filter.interface.ts
@@ -10,7 +10,7 @@ export enum SortMenuActionKey {
 
 export interface AdvancedFilterSortingOptions {
     sortBy?: string;
-    sortDir?: string;
+    sortDir?: 'asc' | 'dsc';
 }
 
 // Note: condition = undefined is only when we add an empty filter rule
--- a/web_ui/src/core/services/urls.ts
+++ b/web_ui/src/core/services/urls.ts
@@ -620,7 +620,7 @@ const TEST_MEDIA_FILTER = (
 
     if (sortingOptions?.sortBy) {
         searchOptionsUrl.set('sort_by', sortingOptions?.sortBy ?? 'score');
-        searchOptionsUrl.set('sort_direction', sortingOptions?.sortDir ?? 'asc');
+        searchOptionsUrl.set('sort_direction', sortingOptions?.sortDir?.toLocaleLowerCase() ?? 'asc');
     }
 
     searchOptionsUrl.set('limit', mediaItemsLoadSize?.toString() ?? '');
--- a/web_ui/src/core/services/urls.ts
+++ b/web_ui/src/core/services/urls.ts
@@ -620,7 +620,7 @@ const TEST_MEDIA_FILTER = (
 
     if (sortingOptions?.sortBy) {
         searchOptionsUrl.set('sort_by', sortingOptions?.sortBy ?? 'score');
-        searchOptionsUrl.set('sort_direction', sortingOptions?.sortDir?.toLocaleLowerCase() ?? 'asc');
+        searchOptionsUrl.set('sort_direction', sortingOptions?.sortDir ?? 'asc');
     }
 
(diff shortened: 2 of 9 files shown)


jpggvilaca (approved):

web_ui/src/core/media/media-filter.interface.ts
```diff
@@ -10,7 +10,7 @@ export enum SortMenuActionKey {
 
 export interface AdvancedFilterSortingOptions {
     sortBy?: string;
-    sortDir?: string;
+    sortDir?: 'asc' | 'dsc';
```
dwesolow, line 13: Ideally we should keep only one `SortDirection` and maybe do it lowercase instead of uppercase?
Do you know why we need to use `dsc` instead of `desc`?

MarkRedeman: `dsc` is what is being used [by the backend](https://github.com/open-edge-platform/geti/blob/main/interactive_ai/libs/iai_core_py/iai_core/repos/project_repo_helpers.py#L32)

I was also thinking of changing `SortDirection` to be lower case, but wasn't sure if this would have any effect on other parts of the code base. So I figured it would be better to do that in a separate PR.

MarkRedeman: To add: it seems like there are some other endpoints, [like jobs](https://github.com/open-edge-platform/geti/blob/main/interactive_ai/services/jobs/app/microservice/job_manager.py#L76) that do use `desc`.
I think the best option here would be to try to keep a consistent usage of `sortDir` as our `SortDirection` enum, and whenever we interact with an API endpoint that we explicitely map it.

Updating the backend to always use `desc` or `dsc` will likely be seen as a backwards compatibility break unless we add a fallback.
@leoll2 do you have any insight into how much effort it would be to update this?

leoll2: We could easily change "dsc" to "desc", which I think is more standard (SQL uses the same notation). As you said, however, that would be a breaking API change, therefore we should support both versions for a while and deprecate the old one. If you think the UI code would significantly benefit from this, feel free to create a ticket for the backend too.

jpggvilaca: @MarkRedeman @leoll2 i dont think we'll release this sprint, so, should we fix this properly on both ends before the next release? If we were to release sooner i'm ok with "fix then improve" but since that is not the case we should take our time. Thoughts?

jpggvilaca (approved):

web_ui/src/core/media/media-filter.interface.ts
MarkRedeman: Let's get this merged and released first. Then look into improving the consistency of the API as a follow up step.

> If you think the UI code would significantly benefit from this, feel free to create a ticket for the backend too

I don't think it will significantly improve things, but it will probably help avoid future issues.

dwesolow (approved):

PR #298, opened 2025-05-26
chore: Update jest config to make it so that some unit tests don't run twice

dwesolow:
## 📝 Description

The issue is that we have "[roots](https://jestjs.io/docs/configuration#roots-arraystring)" and "[projects](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig)" attributes in the jest configs that are responsible in detecting which tests should run. In this commit I remove the "roots" key so we only rely on the recommended for monorepo "[projects](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig)" attribute.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/packages/config/test/jest.config.js
+++ b/web_ui/packages/config/test/jest.config.js
@@ -21,11 +21,5 @@ export default {
     },
     moduleFileExtensions: ['web.js', 'js', 'web.ts', 'ts', 'web.tsx', 'tsx', 'json', 'web.jsx', 'jsx', 'node'],
     resetMocks: false,
-    coverageReporters: ['clover', 'json', 'json-summary'],
-    coverageThreshold: {
-        global: {
-            lines: 75,
-        },
-    },
     globalSetup: path.resolve(dirname, './jest.global.js'),
 };
--- a/web_ui/jest.config.js
+++ b/web_ui/jest.config.js
@@ -4,24 +4,39 @@
 import sharedConfig from '@geti/config/test';
 
 export default {
-    ...sharedConfig,
-    roots: ['<rootDir>/src', '<rootDir>/packages/'],
-    projects: ['<rootDir>', '<rootDir>/packages/*'],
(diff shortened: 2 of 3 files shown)


web_ui/packages/config/test/jest.config.js
```diff
@@ -21,11 +21,5 @@ export default {
     },
     moduleFileExtensions: ['web.js', 'js', 'web.ts', 'ts', 'web.tsx', 'tsx', 'json', 'web.jsx', 'jsx', 'node'],
     resetMocks: false,
-    coverageReporters: ['clover', 'json', 'json-summary'],
```
MarkRedeman, line 24: Can we keep the coverage reporters, but remove the threshold? 
When I was testing things with reporting the coverage the unit tests would fail due to the coverage being calculated per shard instead of combined shards, so I'm fine with removing the threshold in our config.

MarkRedeman (approved):

web_ui/packages/config/test/jest.config.js
dwesolow: Previously I saw a warning that `coverageReporters` does not exists (or sth like that) when running unit test. I discovered it's because it was duplicated in each config and probably jest would not know how to collect and combine coverages from each config. At the moment we have `coverageReporters` in the main config so it should be fine.

image

MarkRedeman: Ok, perhaps it was renamed to something differently. In that case let's remove it and add it back again when we want to enable reporting coverage again.

dwesolow: https://jestjs.io/docs/configuration#coveragereporters-arraystring--string-options. When there is only one `coverageReporters` there is no issue.

jpggvilaca (approved):

PR #341, opened 2025-05-28
ITEP-68021 - background annotations interface

### LINKED ISSUE #689: Holes in annotation shapes
**User story**

Being able to annotate shapes with holes has been a recurring request from various users. An example use case is related to manufacturing, where operators are doing quality assessment of industrial objects during the production process. As part of this use case, the users want to differentiate between the outer and inner ring surface of this object. To achieve this, they need to be able to annotate the outer ring in a 'doughnut' shape. 

Objective of this requirement is to facilitate annotation of objects with holes, so that users can train a model to segment the surface area of such a complex object. This capability will increase the use case coverage by supporting more complex shapes for annotation.

**Requirements**

- Geti shall support annotations of complex shapes with holes as part of instance and semantic segmentation tasks (including segmentation in task chain).
   - Users shall enable hole annotation by adding the 'hole' label (which is predefined by the system) to the label list of their existing project.
   - After doing that, users shall be able to annotate a region with the 'hole' label to mark it as negative space inside an annotated object. 
   - Hole annotation shall be supported across all different segmentation annotation assistants (e.g. SAM, interactive segmentation, etc.).

- Adding/removing/editing hole annotations shall follow the same behaviour as any other annotations.

- Annotation holes should be considered/included in dataset import/export scenarios.

- The SDK shall produce predictions that include holes if applicable.

- Documentation shall be updated accordingly, at minimum with a notion where in the UI users can enable the hole annotation option and how to work with it.

camiloHimura:
## 📝 Description

This PR Updates label interfaces and utilities to support the new background label, gated behind a feature flag.

What’s next:
An upcoming PR will allow users to configure when this label is visible and how it is rendered in the annotator.

Screenshot 2025-05-28 at 15 28 07

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist


web_ui/src/core/projects/services/utils.ts
```diff
@@ -81,6 +82,7 @@ const getRevisitLabel = (label: RevisitLabelDTO, domain: DOMAIN): RevisitLabel =
         behaviour,
         revisitAffectedAnnotations: revisit_affected_annotations,
         isEmpty: is_empty,
+        isBackground: is_background ?? false,
```
ActiveChooN: I can see in `LabelCommon` that `is_background` is already boolean, why do we need nullcheck here?

web_ui/src/core/projects/services/utils.ts
```diff
@@ -225,6 +230,7 @@ const getPerformanceDTO = (performance: Performance): PerformanceDTO => {
 
 const getCommonTaskStructure = (task: TaskDTO, domain: DOMAIN) => {
     const labels = (task.labels ?? []).map((label) => {
+        //const labels = initialLabels.map((label) => {
```
ActiveChooN: leftovers?

ActiveChooN (dismissed):
LGTM

web_ui/src/core/projects/services/utils.ts
camiloHimura: It hasn't been implemented on the backend yet, so this is just a fail-safe

web_ui/src/core/projects/services/utils.ts
camiloHimura: yes, removed

web_ui/src/core/annotations/services/utils.ts
```diff
@@ -198,6 +198,7 @@ export const getAnnotation = (annotation: AnnotationDTO, projectLabels: Label[],
                 hotkey,
                 behaviour,
                 isEmpty,
+                isBackground,
```
MarkRedeman: Could we include `isBackground` inside of `behaviour?

ActiveChooN (dismissed):

web_ui/src/core/annotations/services/utils.ts
camiloHimura: sure, added in my last commit https://github.com/open-edge-platform/geti/pull/341/files#diff-25ed967fbd7e0f6f997bb83ee6931509e13265a84494aa307a10a8a385fd8477R61

web_ui/src/core/labels/utils.ts
```diff
@@ -61,9 +61,31 @@ export const isAnomalous = <T extends { behaviour: LABEL_BEHAVIOUR }>(label: T):
...
 
-export const isEmptyLabel = <T extends { isEmpty: boolean }>(label: T): boolean => label.isEmpty;
+export const isBackground = <T extends { behaviour: LABEL_BEHAVIOUR }>(label: T): boolean => {
+    return Boolean(label.behaviour & LABEL_BEHAVIOUR.BACKGROUND);
+};
+
+interface EmptyOrBackground {
+    isEmpty: boolean;
+    isBackground: boolean;
+    behaviour: LABEL_BEHAVIOUR;
```
jpggvilaca, line 70: qq: why do we need `behavior` here?

web_ui/src/core/labels/utils.ts
```diff
@@ -100,6 +122,10 @@ export const getBehaviourFromDTO = (
         return LABEL_BEHAVIOUR.GLOBAL;
     }
 
+    if (isKeypointDetection(domain)) {
```
jpggvilaca: could you elaborate on this one?

web_ui/src/core/labels/label.interface.ts
```diff
@@ -56,6 +57,8 @@ export enum LABEL_BEHAVIOUR {
     // Used to guarantee that when an anomalous annotation is added, its roi is
     // also given a global anomalous label
     ANOMALOUS = 1 << 4,
+
+    BACKGROUND = 1 << 5,
```
jpggvilaca, line 60: since we're adding `BACKGROUND`, should we consider (on a different PR perhaps) adding `EMPTY` as well?

web_ui/src/core/labels/utils.ts
camiloHimura: It's validated at line 78

web_ui/src/core/labels/utils.ts
camiloHimura: Sorry, I got it mixed up with another PR’s config. updated

web_ui/src/core/labels/label.interface.ts
camiloHimura: That's a good point, but since behaviours isn't sent to the server, we could consider revisiting it to simplify the implementation. Let's bring this up in the next tech talk.

jpggvilaca (dismissed):

web_ui/src/core/labels/label.interface.ts
MarkRedeman: > since we're adding `BACKGROUND`, should we consider (on a different PR perhaps) adding `EMPTY` as well?

That behaviour is handled via `EXCLUSIVE`, when a label has this behaviour the annotator won't allow other labels at the same global scope.
So for instance for all local tasks there can be only 1 empty global annotation, but if you have detection -> classification, then at the classification level there can be multiple detection annotations that have an empty classification label.

web_ui/src/core/labels/label.interface.ts
```diff
@@ -9,6 +9,7 @@ interface LabelCommon {
     readonly hotkey?: string;
     readonly behaviour: LABEL_BEHAVIOUR;
     readonly isEmpty: boolean;
+    readonly isBackground: boolean;
```
MarkRedeman: We should no longer need this field with the addition of `BACKGROUND` to `LABEL_BEHAVIOUR`.
For any code that needs to check if a label is a background label we should be able to do `isBackgroundLabel(label)` which returns it based on the label behaviour.

camiloHimura: ok, but should we keep it for `LabelCommon` DTO?

web_ui/src/core/projects/services/utils.ts
```diff
@@ -235,23 +236,41 @@ const getCommonTaskStructure = (task: TaskDTO, domain: DOMAIN) => {
...
         title: task.title,
         domain,
-        labels: emptyLabel === undefined ? labelsSortedByStructure : [...labelsSortedByStructure, emptyLabel],
+        labels: [...labelsSortedByStructure, ...getEmptyAndBackgroundLabels(labels)],
     };
 
     return commonStructure;
 };
 
+const getEmptyAndBackgroundLabels = (labels: Label[]) => {
```
MarkRedeman, line 255: What about,
```
return labels.filter(label => isBackgroundLabel(label) || isEmptyLabel(label));
```

web_ui/src/core/labels/label.interface.ts
MarkRedeman: I think we should remove it there as well. So essentially we should be able to determine all of the label's behaviour from its behaviour field.
Tbh I'm not entirely sure why we also have the `isEmpty` field while that can be derived from `behaviour` as well. That might be something we can look into in the future?

camiloHimura: removed

web_ui/src/core/projects/services/utils.ts
camiloHimura: updated

jpggvilaca (approved):

ActiveChooN (approved):

PR #351, opened 2025-05-29
ITEP-68202 decreasing cpu requests

jcchr:
## 📝 Description

The PR decreases cpu resource requests settings for several components and applies reduction of requests for resources for all components - without a need of chosing a profile. The goal is to allow installing platform on machines with 16 CPU cores - as now - due to requested number of CPUs - it is not possible,

JIRA: ITEP-68202, ITEP-68877

-->

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

To test changes - install the platform in both profiles - standard and lite - and check the settings for the cpu requests. In the "lite" profile it should be set to far less than 16 CPU.

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/interactive_ai/services/inference_gateway/chart/templates/deployment.yaml
+++ b/interactive_ai/services/inference_gateway/chart/templates/deployment.yaml
@@ -47,7 +47,11 @@ spec:
           {{- toYaml .Values.securityContext | nindent 12 }}
           env:
             - name: GOMAXPROCS
-              value: "{{ .Values.resources.requests.cpu }}"
+              {{- if (eq .Values.global.installation_profile "lite") }}
+              value: "{{ .Values.liteResources.requests.cpu }}"
+              {{- else }}
+              value: "{{ .Values.standardResources.requests.cpu }}"
+              {{- end }}
             - name: GOMEMLIMIT
               value: "{{ .Values.global.mem_limit }}"
             - name: WORKDIR
--- a/interactive_ai/services/inference_gateway/chart/templates/deployment.yaml
+++ b/interactive_ai/services/inference_gateway/chart/templates/deployment.yaml
@@ -125,7 +125,11 @@ spec:
             - configMapRef:
                 name: {{ .Release.Namespace }}-s3-bucket-names
           resources:
-            {{- toYaml .Values.resources | nindent 12 }}
+            {{- if (eq .Values.global.installation_profile "lite") }}
+              {{ toYaml .Values.liteResources | nindent 12 }}
+            {{- else }}
+              {{ toYaml .Values.standardResources | nindent 12 }}
+            {{- end }}
           volumeMounts:
             - mountPath: {{ .Values.mountPath }}
               name: data-storage
--- a/interactive_ai/services/inference_gateway/chart/templates/deployment.yaml
+++ b/interactive_ai/services/inference_gateway/chart/templates/deployment.yaml
@@ -47,11 +47,7 @@ spec:
           {{- toYaml .Values.securityContext | nindent 12 }}
           env:
             - name: GOMAXPROCS
-              {{- if (eq .Values.global.installation_profile "lite") }}
-              value: "{{ .Values.liteResources.requests.cpu }}"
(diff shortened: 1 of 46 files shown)


interactive_ai/services/inference_gateway/chart/templates/deployment.yaml
```diff
@@ -47,7 +47,11 @@ spec:
           {{- toYaml .Values.securityContext | nindent 12 }}
           env:
             - name: GOMAXPROCS
-              value: "{{ .Values.resources.requests.cpu }}"
+              {{- if (eq .Values.global.installation_profile "lite") }}
+              value: "{{ .Values.liteResources.requests.cpu }}"
+              {{- else }}
+              value: "{{ .Values.standardResources.requests.cpu }}"
+              {{- end }}
```
AlbertvanHouten: This env var does not work with fractional CPU.  In this scenario, it should be set to 1 for lite profile.

interactive_ai/services/media/chart/templates/deployment.yaml
```diff
@@ -47,7 +47,11 @@ spec:
           {{- toYaml .Values.securityContext | nindent 12 }}
           env:
             - name: GOMAXPROCS
-              value: "{{ .Values.resources.requests.cpu }}"
+              {{- if (eq .Values.global.installation_profile "lite") }}
+              value: "{{ .Values.liteResources.requests.cpu }}"
+              {{- else }}
+              value: "{{ .Values.standardResources.requests.cpu }}"
+              {{- end }}
```
AlbertvanHouten: same here

interactive_ai/workflows/geti_domain/common/jobs_common/k8s_helpers/k8s_resources_calculation.py
```diff
@@ -26,6 +26,10 @@
         "requests": {"cpu": "2", "memory": "20GB"},
         "limits": {"cpu": "100", "memory": "50GB"},
     },
+    "LITE": {
+        "requests": {"cpu": "2", "memory": "1GB"},
```
AlbertvanHouten: If possible, we should increase this memory value. Most training jobs will not work with only 1 GB, and if the system is already fully occupied, it will result in crashes.

jcchr: It is expected, that installation on environments with lower resources may lead to some limitations - like inability to execute training for some models/datasets. In such cases we will display information about the cause of it - to inform a user that an error was caused by a lack of resources. However decreasing this limit allows installing the platform on such machines at all - and as still some training jobs will run - a user will see Geti in action.

AlbertvanHouten: This value doesn't have any effect on installation though, it only affects scheduling of jobs. 

I would strongly suggest doing some testing by also setting the limit to 1 GB and testing which jobs will still work. If most of the jobs end up failing there is not really any point in putting such a low value IMO.

leoll2: > In such cases we will display information about the cause of it - to inform a user that an error was caused by a lack of resources.

I have doubts about the technical feasibility of this approach, let's discuss it offline. UX is also degraded as it informs the user very late about insufficient HW resources, potentially wasting a lot of time.

leoll2: For reference: we agreed to tune the memory "requests" value based on the memory consumption of the most popular model (ATSS) for the most popular task (detection) when training on 12 images. The value is 3GB, determined empirically.

leoll2 (approved):

piotrgrubicki (approved):

dwesolow (approved):

PR #357, opened 2025-05-29
ITEP-59431 Disable telemetry for lite installation

dwesolow:
## 📝 Description

We don't want to provide telemetry for the geti lite.

~~`product_info` response is static so I added `staleTime: Infinity, gc: Infinity`.
More context:
I wanted to have `installation-profile` to be added to `deployment_config`. However, I got informed that adding that to `deployment_config` instead of `product_info` would cause duplicating values (it'd be in two config maps - impt-configuration i w deployment-config).~~

**Edit: I removed implementation related to adding `installation-profile` to `product_info` and I exposed small hook without implementation that will allow me to implement the UI part and once BE is finalized I will do real hook implementation (either using ff or deployment config, will not matter).**

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/analytics/analytics-provider.component.tsx
+++ b/web_ui/src/analytics/analytics-provider.component.tsx
@@ -9,7 +9,6 @@ import { MeterProvider, PeriodicExportingMetricReader } from '@opentelemetry/sdk
 
 import { usePlatformUtils } from '../core/platform-utils/hooks/use-platform-utils.hook';
 import { useEventListener } from '../hooks/event-listener/event-listener.hook';
-import { MissingProviderError } from '../shared/missing-provider-error';
 import { createPeriodicMetricExporter, initializeMetrics } from './metrics';
 import { initializeTracing } from './traces';
 import { SERVICE_DEFAULT_INFO } from './utils';
@@ -88,8 +87,7 @@ const AnalyticsProvider = ({ children }: AnalyticsProviderProps): JSX.Element =>
             meter: meterProviderRef.current?.getMeter('default', SERVICE_DEFAULT_INFO.serviceVersion),
             getTrace: tracingRef.current?.getTrace,
         }),
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-        [meterProviderRef.current, tracingRef.current]
+        []
     );
 
     return <AnalyticsContext.Provider value={value}>{children}</AnalyticsContext.Provider>;
@@ -99,7 +97,10 @@ const useAnalytics = (): AnalyticsContextProps => {
     const context = useContext(AnalyticsContext);
 
     if (context === undefined) {
-        throw new MissingProviderError('useAnalytics', 'AnalyticsProvider');
+        return {
+            meter: undefined,
+            getTrace: undefined,
+        };
     }
 
     return context;
--- a/web_ui/src/analytics/analytics-provider.component.tsx
+++ b/web_ui/src/analytics/analytics-provider.component.tsx
@@ -7,7 +7,7 @@ import { useApplicationServices } from '@geti/core/src/services/application-serv
 import { Meter } from '@opentelemetry/api';
 import { MeterProvider, PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics';
 
-import { usePlatformUtils } from '../core/platform-utils/hooks/use-platform-utils.hook';
+import { useWorkflowId } from '../core/platform-utils/hooks/use-platform-utils.hook';
 import { useEventListener } from '../hooks/event-listener/event-listener.hook';
 import { createPeriodicMetricExporter, initializeMetrics } from './metrics';
 import { initializeTracing } from './traces';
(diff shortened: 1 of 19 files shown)


web_ui/src/core/platform-utils/dto/utils.interface.ts
```diff
@@ -12,6 +12,8 @@ export enum GPUProvider {
     NVIDIA = 'nvidia',
 }
 
+export type InstallationMode = 'standard' | 'lite';
```
MarkRedeman: This feels like the same as our serving mode, would it be possible to include it there so we would have `saas`, `on-prem`, `lite`?

web_ui/src/pages/about-page/about-page.component.tsx
```diff
@@ -7,16 +7,15 @@ import { ActionButton, Flex, Heading, Link as SpectrumLink, Text, View } from '@
...
 import { useIsSaasEnv } from '../../hooks/use-is-saas-env/use-is-saas-env.hook';
 import { LicenseModal } from '../../shared/components/license-modal/license-modal.component';
 import { PageLayout } from '../../shared/components/page-layout/page-layout.component';
 
 import classes from './about-page.module.scss';
 
 const AboutPage = (): JSX.Element => {
-    const { useProductInfo } = usePlatformUtils();
-    const { isPending, data } = useProductInfo();
+    const { data } = useProductInfo();
```
MarkRedeman: Do we have a suspense boundary above `AboutPage`? Otherwise since `useProductInfo` is a suspense query it could trigger reloading a lot of our components.
Perhaps we can extract a `AboutPageContent` in this same component and do something like,
```
        <PageLayout breadcrumbs={[{ id: 'about-id', breadcrumb: 'About' }]}>
            <Suspense fallback={<Loading />}><AboutPageContent /></Suspense>
        </PageLayout>
```

web_ui/src/routes/app-routes.component.tsx
```diff
@@ -104,7 +104,7 @@ const AppProviders = (): JSX.Element => {
 
     return (
         <Suspense fallback={<IntelBrandedLoading />}>
-            <AnalyticsProvider>
+            <InstallationModeProvider>
```
MarkRedeman, line 107: This would require us to wait for the product info to be loaded before we can start showing/loading the actual page the user is looking for.
So we'd wait for the following before loading the normal routes:
- `/deployment-config.json` - informing us about which api endpoints to use and the serving mode (can include lite mode?)
- `/api/v1/feature_flags` - telling us which features to enable/disable (we can opt out of analytics by using a feature flag instead of lite mode?)
- User info from OIDC
- `/api/v1/product_info`

I think that adding `product_info` to this group is a bit too much as the same info could be taken from the deployment config or feature flags as well.

Could we do one of the following:
1. Look into disabling telemetry/analytics by adding a `FEATURE_FLAG_ANALYTICS` option, that users can enable/disable as they see fit
2. Look into adding `lite` as a possible option for the serving mode from `/deployment-config.json`

Both of these could be good solutions as we already need to wait for this data before showing the application.

web_ui/src/analytics/analytics-provider.component.tsx
```diff
@@ -99,10 +96,18 @@ const useAnalytics = (): AnalyticsContextProps => {
...
+        };
     }
 
     return context;
 };
 
-export { AnalyticsProvider, useAnalytics };
+const useIsAnalyticsEnabled = (): boolean => {
+    // TODO: Implement a more robust check for analytics enablement
+    return true;
```
pplaskie, line 110: :D

romanowska (approved):

web_ui/src/routes/app-routes.component.tsx
pplaskie: Adding 'lite' as an option to the serving mode will be very helpful either way. It's not all of the work needed for lite versione, is it?

dwesolow: Totally agree (as we discussed on daily). I reverted this change not to block this PR and we will adjust the implementation once BE is finalized.

web_ui/src/pages/about-page/about-page.component.tsx
dwesolow: It's no longer valid (but it'd not be the case here because `InstallationModeProvider` would get the product info first and then in about we already would have it so `Suspense` would not be needed).

web_ui/src/core/platform-utils/dto/utils.interface.ts
dwesolow: To discuss whether we want to update `deployment_config` or `feature_flags`.
CC @jcchr

web_ui/src/routes/app-routes.component.tsx
dwesolow: > Adding 'lite' as an option to the serving mode will be very helpful either way. It's not all of the work needed for lite versione, is it?

I'm aware only of these two tickets for the UI:
https://jira.devtools.intel.com/browse/ITEP-67071
https://jira.devtools.intel.com/browse/ITEP-68094

pplaskie: Ok, not much. But still will be useful

pplaskie (approved):

PR #359, opened 2025-05-29
ITEP-31796 Investigate useDatasetImportQueries for unnecessary use of useEffects

romanowska:
## 📝 Description

Investigate useDatasetImportQueries for unnecessary use of useEffects

JIRA: ITEP-31796

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/core/datasets/hooks/use-dataset-import-queries.hook.ts
+++ b/web_ui/src/core/datasets/hooks/use-dataset-import-queries.hook.ts
@@ -1,7 +1,7 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
 
 import { useApplicationServices } from '@geti/core/src/services/application-services-provider.component';
 import { useMutation, UseMutationResult, useQuery, UseQueryResult } from '@tanstack/react-query';
@@ -142,8 +142,6 @@ export const useDatasetImportQueries = (): UseDatasetImportQueries => {
         enabled = true,
         ...intervalHandlers
     }) => {
-        const intervalHandlersRef = useRef(intervalHandlers);
-
         const query = useQuery<JobPrepareDatasetImportNewProjectStatus, AxiosError>({
             queryKey: QUERY_KEYS.PREPARE_DATASET_STATUS_JOB_KEY(data),
             queryFn: () => service.prepareDatasetImportNewProjectJob(data),
@@ -152,23 +150,13 @@ export const useDatasetImportQueries = (): UseDatasetImportQueries => {
         });
 
         useEffect(() => {
-            intervalHandlersRef.current = intervalHandlers;
-        }, [intervalHandlers]);
-
-        useEffect(() => {
-            if (!enabled || !query.isSuccess) {
-                return;
+            if (!enabled) return;
+            if (query.isSuccess) {
+                getIntervalJobHandlers(intervalHandlers)(query.data);
+            } else if (query.isError && query.error !== null) {
+                intervalHandlers.onError(query.error);
             }
-            getIntervalJobHandlers(intervalHandlersRef.current)(query.data);
-        }, [enabled, query.isSuccess, query.data]);
-
-        useEffect(() => {
-            if (!query.isError || query.error === null) {
-                return;
-            }
-
-            intervalHandlersRef.current.onError(query.error);
-        }, [query.isError, query.error]);
+        }, [enabled, query.data, query.isSuccess, query.isError, query.error, intervalHandlers]);
 
         return query;
     };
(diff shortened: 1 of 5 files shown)


pplaskie (dismissed):

dwesolow:
IMHO it does not change anything regarding to the initial plan of this ticket (the plan was to investigate the whole dataset import feature and check if it's possible to remove some `useEffect`s). I preferred the previous way of having multiple `useEffect`s where each `useEffect` had its own responsibility and array of dependencies was minimal.

romanowska:
> IMHO it does not change anything regarding to the initial plan of this ticket (the plan was to investigate the whole dataset import feature and check if it's possible to remove some `useEffect`s). I preferred the previous way of having multiple `useEffect`s where each `useEffect` had its own responsibility and array of dependencies was minimal.

First of all - as you can see I got rid of the first useEffect entirely because as I checked we don't need to use the handlers in reference because it doesn't change anything.

About success/error handling - I decided to give it a try since in my opinion it improves readability and in any case you would go to one of the useEffects, so one still would be called.

About investigation of use of this hook in our app, I did check and all of them are used and needed.

pplaskie:
@dwesolow There is a useEffect removed and merging those two which left increasing readability IMHO. And those functions are called only when necessary, it didn't introduce any unneded operations.

web_ui/src/core/datasets/hooks/use-dataset-import-queries.hook.ts
```diff
@@ -152,23 +150,13 @@ export const useDatasetImportQueries = (): UseDatasetImportQueries => {
         });
 
         useEffect(() => {
-            intervalHandlersRef.current = intervalHandlers;
-        }, [intervalHandlers]);
-
```
MarkRedeman: I'm wondering if this effect isn't one of those "necessary evils" we'd need to have. With the change from this PR if `intervalHandlers` is changed (because of a rerender) then the `useEffect` would re trigger which could incorrectly trigger the job handlers

romanowska: Yes, that's what I was worried about too after discussing it with Copilot but I manually tested it thoroughly and it works like a charm so.. I don't know..

dwesolow:
> > IMHO it does not change anything regarding to the initial plan of this ticket (the plan was to investigate the whole dataset import feature and check if it's possible to remove some `useEffect`s). I preferred the previous way of having multiple `useEffect`s where each `useEffect` had its own responsibility and array of dependencies was minimal.
> 
> First of all - as you can see I got rid of the first useEffect entirely because as I checked we don't need to use the handlers in reference because it doesn't change anything.
> 
> About success/error handling - I decided to give it a try since in my opinion it improves readability and in any case you would go to one of the useEffects, so one still would be called.
> 
> About investigation of use of this hook in our app, I did check and all of them are used and needed.

Example where it's not used and might be removed:
image

The usage of  `ref` there was intentional to keep stable reference. At the moment, when added `intervalHandlers` as dependency, `useEffect` will be triggered every time parent component rerenders (so if the hook has `isSuccess: true` it might be called not only once when it changed from false to true but x times because of parent rerendering).

romanowska:
> > > IMHO it does not change anything regarding to the initial plan of this ticket (the plan was to investigate the whole dataset import feature and check if it's possible to remove some `useEffect`s). I preferred the previous way of having multiple `useEffect`s where each `useEffect` had its own responsibility and array of dependencies was minimal.
> > 
> > 
> > First of all - as you can see I got rid of the first useEffect entirely because as I checked we don't need to use the handlers in reference because it doesn't change anything.
> > About success/error handling - I decided to give it a try since in my opinion it improves readability and in any case you would go to one of the useEffects, so one still would be called.
> > About investigation of use of this hook in our app, I did check and all of them are used and needed.
> 
> Example where it's not used and might be removed: image
> 
> The usage of `ref` there was intentional to keep stable reference. At the moment, when added `intervalHandlers` as dependency, `useEffect` will be triggered every time parent component rerenders (so if the hook has `isSuccess: true` it might be called not only once when it changed from false to true but x times because of parent rerendering).

Yes, in two places it doesn't require the handlers but for example in DatasetImportJobStatus we use it with several handlers.
Screenshot 2025-05-30 at 09 36 05

dwesolow:
> > > > IMHO it does not change anything regarding to the initial plan of this ticket (the plan was to investigate the whole dataset import feature and check if it's possible to remove some `useEffect`s). I preferred the previous way of having multiple `useEffect`s where each `useEffect` had its own responsibility and array of dependencies was minimal.
> > > 
> > > 
> > > First of all - as you can see I got rid of the first useEffect entirely because as I checked we don't need to use the handlers in reference because it doesn't change anything.
> > > About success/error handling - I decided to give it a try since in my opinion it improves readability and in any case you would go to one of the useEffects, so one still would be called.
> > > About investigation of use of this hook in our app, I did check and all of them are used and needed.
> > 
> > 
> > Example where it's not used and might be removed: image
> > The usage of `ref` there was intentional to keep stable reference. At the moment, when added `intervalHandlers` as dependency, `useEffect` will be triggered every time parent component rerenders (so if the hook has `isSuccess: true` it might be called not only once when it changed from false to true but x times because of parent rerendering).
> 
> Yes, in two places it doesn't require the handlers but for example in DatasetImportJobStatus we use it with several handlers. Screenshot 2025-05-30 at 09 36 05

Ok but then it can still be removed not to have `noop`, right?

pplaskie (approved):

dwesolow (approved):

PR #373, opened 2025-06-02
Remove default range jobs filtering and disable filtering scheduled jobs

pplaskie:
## 📝 Description

1) The default range filter is not required. My changes have removed the default range and now allow users to clear the filter directly in the calendar component.
2) There is a known issue with scheduled jobs, as they do not have a start time. I removed this jobs filtering in this PR. We will revert this once filtering by creation time is supported:
- scheduled jobs are not filtered
- calendar on scheduled jobs is disabled
- filtering by creation time should be enable during next sprint

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
```diff
@@ -83,18 +113,26 @@ export const JobsDialog = ({ isFullScreen, onClose, setIsFullScreen }: JobsDialo
...
+    );
+
     const {
         numberOfRunningJobs,
         numberOfFinishedJobs,
         numberOfScheduledJobs,
         numberOfCancelledJobs,
         numberOfFailedJobs,
-    } = data?.pages?.at(0)?.jobsCount ?? DEFAULT_JOBS_COUNT;
+    } = jobsCountRef.current;
```
MarkRedeman, line 135: One part that worries me slightly is that the react docs [recommend not to read/write from refs during renders](https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs).
In this case this probably is fine since we are setting the ref right above at line 121. But let's keep an eye out in case we make any more changes to this component.

MarkRedeman (approved):

web_ui/src/shared/components/header/jobs-management/jobs-dialog.component.tsx
pplaskie: Thanks for warning but It's just "unpacking" the attributes from object so shouldn't be problematic. And moreover, it's most likely for one sprint only - after that I'll return old code and use new attribute to filter so there will be no refs.

dwesolow (approved):

PR #376, opened 2025-06-03
ITEP-68215 - Update configuration for model entity

maxxgx:
## 📝 Description

Add a dictionary to the model configuration class containing the revamped configuration. This configuration is not intended to be modified, as its main purpose is for retrieval and visualization.

The configuration dictionary is passed as job payload to be used in the training job, as this is where the models are created.

JIRA: ITEP-68215

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


AlbertvanHouten:
If I understand correctly, we are now duplicating this configuration for every model of a specific architecture type. The values will also never change. Couldn't we make it a part of the model manifest instead and load it from there to prevent this duplication?

leoll2:
@AlbertvanHouten this is a snapshot of the configuration used to train a specific model. It can't be modified later because it's part of the model history, but the values can differ from the default ones found in the model template if the user chooses custom training settings.

maxxgx:
Manually tested that all new endpoints are working

interactive_ai/services/director/app/storage/mappers/project_configuration_mapper.py
```diff
@@ -17,6 +17,11 @@ def forward(instance: TaskConfig) -> dict:
         doc["task_id"] = IDToMongo.forward(instance.task_id)
         return doc
 
+    @staticmethod
+    def backward(instance: dict) -> TaskConfig:
+        instance["task_id"] = str(IDToMongo.backward(instance["task_id"]))  # Convert ObjectId back to string
```
leoll2, line 22: Why is it string?

interactive_ai/services/director/app/service/job_submission/train.py
```diff
@@ -77,6 +78,13 @@ def prepare_data(  # type: ignore  # noqa: PLR0913
 
         keep_mlflow_artifacts = FeatureFlagProvider.is_enabled(FeatureFlag.FEATURE_FLAG_RETAIN_TRAINING_ARTIFACTS)
 
+        full_training_configuration = ConfigurationService.get_full_training_configuration(
```
leoll2: Should this be behind FF?

interactive_ai/services/director/app/service/job_submission/job_creation_helpers.py
```diff
@@ -155,6 +158,7 @@ def create_payload(self) -> dict:
             "max_number_of_annotations": self.max_number_of_annotations,
             "reshuffle_subsets": self.reshuffle_subsets,
             "keep_mlflow_artifacts": self.keep_mlflow_artifacts,
+            "hyperparameters": self.training_configuration.hyperparameters.model_dump(),
```
leoll2: Should this be behind FF?

interactive_ai/services/director/app/storage/mappers/project_configuration_mapper.py
maxxgx: Pydantic models does not support `ID` type

interactive_ai/services/director/app/service/job_submission/job_creation_helpers.py
maxxgx: Technically yes, but job payload needs to match flyte workflow arguments so it might be necessary to add this. I'm testing whether putting under FF still works

leoll2: Including the "hyperparameters" unconditionally is not wrong per se, but it is a potential source of errors if the call to `model_dump()` fails for any reason even if the FF is disabled. In other words, it could introduce regressions while the feature is still under development. But if you test that training works in every PR, then it should be okay.

maxxgx: I verified there are no issue. It's now behind FF

interactive_ai/services/director/app/service/job_submission/train.py
maxxgx: Done

leoll2 (approved):

PR #389, opened 2025-06-04
Fix charts templating: remove circular dependency between Makefiles a…

mpryahin:
## 📝 Description

Fix charts templating: remove circular dependency between Makefiles and Helm charts

Restructured how context is passed to chart templates to break the loop where Makefiles provided rendering context, while also depending on variables from Helm charts. Also enabled components that are not packaged as Helm charts to inherit the Makefile.shared build logic.

Previously, the `name` field in a component's `Chart.yaml` served as the source of truth. Moreover, component artifacts should not be used as the place to store component metadata as today it’s a Helm chart, tomorrow it could be something else.

This approach made it impossible to reuse the build logic defined in `Makefile.shared` for components that are not packaged as helm charts.

Additionally, the makefile logic read this value from `Chart.yaml` and, after some transformations, used it as context to resolve placeholders within `Chart.yaml`, introducing a circular dependency.

This PR makes the component's Makefile the single source of truth for component metadata.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/interactive_ai/services/media/Makefile
+++ b/interactive_ai/services/media/Makefile
@@ -3,5 +3,6 @@
 
 include ../../../Makefile.shared-go
 
+COMPONENT_NAME := media-ms
 DOCKER_BUILD_CONTEXT := --build-context iai_core=$(CWD)/interactive_ai/libs/iai_core_go
 DOCKER_IMAGE_DESCRIPTION := Media Service for Geti
--- a/Makefile.shared
+++ b/Makefile.shared
@@ -82,9 +82,8 @@ publish-image:
 build-chart:
 	mkdir -p ${CHART_BUILD_DIR}
 	cp -r ${CHART_DIR}/* ${CHART_BUILD_DIR}/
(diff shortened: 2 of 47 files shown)


AlbertvanHouten (approved):

interactive_ai/services/media/Makefile
```diff
@@ -3,5 +3,6 @@
 
 include ../../../Makefile.shared-go
 
+COMPONENT_NAME := media-ms
```
leoll2, line 6: The other services don't have the `-ms` suffix. How much effort would it be to remove it?

leoll2 (approved):

interactive_ai/services/media/Makefile
mpryahin: Thanks, Leo! I will create a ticket to unify all component names as I also noticed some components are named in a snake case notation like `weights_downloader`. Regarding efforts - we'll have to configure the container registry.

jcchr (approved):

MarkRedeman (approved):

PR #395, opened 2025-06-05
Redirect user based on local storage instead of user settings

MarkRedeman:
## 📝 Description

Previously we used the user's global settings to determine which organization we wanted to redirect to. This has the disadvantage that we need to wait for loading these settings until we can show the "critical path" of the UI (e.g. the annotator, or project page).
As an alternative this PR uses localstorage to instead store the selected organization. This makes loading the page faster as we don't need to wait for this query to load.

In a follow up PR I'd like to move this redirect logic from the `useEffect` inside of `useSelectedOrganization` into a redirect route inside of our `app-routes` file. This redirect should then only be rendered when the user navigates to say https://geti.dev instead of https://geti.dev/organizations/859a4dc7-a055-47a1-8312-edc7c6b1968c.
Making this follow up change should fix a bug where a user may have more than 1 organization, tries to open a specific url but is redirected because the last time they used geti was on their other organization.

## ✨ Type of Change

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/core/organizations/hook/use-selected-organization.hook.ts
+++ b/web_ui/src/core/organizations/hook/use-selected-organization.hook.ts
@@ -17,9 +17,6 @@ import { useNotification } from '../../../notification/notification.component';
 import { isActiveOrganization, isUserActivatedInOrg, isUserInvitedInOrg } from '../../../routes/organizations/util';
 import { LOCAL_STORAGE_KEYS } from '../../../shared/local-storage-keys';
 import { hasEqualId } from '../../../shared/utils';
-import { GENERAL_SETTINGS_KEYS } from '../../user-settings/dtos/user-settings.interface';
-import { useUserGlobalSettings } from '../../user-settings/hooks/use-global-settings.hook';
-import { getSettingsOfType } from '../../user-settings/utils';
 
 const removeLastForwardSlash = (text: string) => text.replace(/\/$/, '');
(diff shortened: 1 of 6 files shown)


dwesolow:
> In a follow up PR I'd like to move this redirect logic from the useEffect inside of useSelectedOrganization into a redirect route inside of our app-routes file.

https://jira.devtools.intel.com/browse/CVS-163754

web_ui/src/core/organizations/hook/use-selected-organization.hook.ts
```diff
@@ -111,10 +114,8 @@ export const useSelectedOrganization = () => {
                 return;
             }
 
-            userGlobalSettings.saveConfig({
-                ...userGlobalSettings.config,
-                [GENERAL_SETTINGS_KEYS.CHOSEN_ORGANIZATION]: { value: newOrganization.id },
```
dwesolow, line 116: `CHOSEN_ORGANIZATION` can be removed from `GENERAL_SETTINGS_KEYS`, right?

MarkRedeman: Yep removed

dwesolow (approved):

jpggvilaca (approved):

PR #399, opened 2025-06-05
feat: Move reshuffle option to be next to train from scratch

dwesolow:
## 📝 Description

`Reshuffle subsets` should be only available when train from scratch is enabled.

image

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/project-details/components/project-models/train-model-dialog/use-train-model-state.hook.ts
+++ b/web_ui/src/pages/project-details/components/project-models/train-model-dialog/use-train-model-state.hook.ts
@@ -104,6 +104,14 @@ export const useTrainModelState = () => {
         setSelectedModelTemplateId(getActiveModelTemplateId(models, algorithms, newTask.id));
     };
 
+    const handleTrainFromScratchChange = (newTrainFromScratch: boolean): void => {
+        setTrainFromScratch(newTrainFromScratch);
+
+        if (newTrainFromScratch === false) {
+            setIsReshufflingSubsetsEnabled(false);
+        }
+    };
+
     return {
         isBasicMode,
         openAdvancedSettingsMode,
@@ -120,6 +128,6 @@ export const useTrainModelState = () => {
         changeReshufflingSubsetsEnabled: setIsReshufflingSubsetsEnabled,
         configParameters,
         trainFromScratch,
-        changeTrainFromScratch: setTrainFromScratch,
+        changeTrainFromScratch: handleTrainFromScratchChange,
     } as const;
 };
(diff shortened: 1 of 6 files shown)


web_ui/src/pages/project-details/components/project-models/train-model-dialog/use-train-model-state.hook.ts
```diff
@@ -104,6 +104,14 @@ export const useTrainModelState = () => {
         setSelectedModelTemplateId(getActiveModelTemplateId(models, algorithms, newTask.id));
     };
 
+    const handleTrainFromScratchChange = (newTrainFromScratch: boolean): void => {
+        setTrainFromScratch(newTrainFromScratch);
+
+        if (newTrainFromScratch === false) {
```
jpggvilaca, line 110:
```suggestion
        setIsReshufflingSubsetsEnabled(newTrainFromScratch);
```

jpggvilaca (approved):
is there an unit test covering this case already?

dwesolow:
> is there an unit test covering this case already?

They will be implemented in https://jira.devtools.intel.com/browse/ITEP-32009

web_ui/src/pages/project-details/components/project-models/train-model-dialog/use-train-model-state.hook.ts
dwesolow: I want to update `isReshufflingSubsetEnabled` to `false` only when `trainFromScratch` is `false`. I don't want to enable it always when setting `trainFromScratch` to `true`.

jpggvilaca: But isnt this a checkbox? IF so we always want to update both state values, or maybe we dont even need this one can can use `trainFromScratch` directly?

dwesolow: No, we want to update both only when `trainFromScratch` is `false` because then reshuffling subsets feature does not make any impact. When `trainFromScratch` is `true` it does not mean that user wants to enable reshuffling subsets (BTW this feature with this logic already exists in the previous implementation of training flow).

jpggvilaca: Ah ic ic. Thanks!

camiloHimura (approved):

PR #416, opened 2025-06-09
Changed workspace permissions - part 1

pplaskie:
## 📝 Description

First part of changed permissions. Work is not done, will be continued.
With switched off FEATURE_FLAG_WORKSPACE_ACTIONS everything should work as it is.
You can find permission description under the task (but implementation didn't cover all the cases for now) :  ITEP-67427

Next steps:
* edition of permission 
* default workspace at the start of tabs list
* workspace deletion
* name edition (fix a bug)

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist


web_ui/src/pages/project-details/components/project-sidebar/project-sidebar.component.tsx
```diff
@@ -31,8 +32,9 @@ interface ProjectSidebarProps {
 
 export const ProjectSidebar = ({ project }: ProjectSidebarProps): JSX.Element => {
     const datasetIdentifier = useDatasetIdentifier();
+    const { FEATURE_FLAG_WORKSPACE_ACTIONS } = useFeatureFlags();
     const canEditProjectName = useCheckPermission(
-        [OPERATION.PROJECT_NAME_EDITION],
+        [FEATURE_FLAG_WORKSPACE_ACTIONS ? OPERATION_NEW.PROJECT_NAME_EDITION : OPERATION_OLD.PROJECT_NAME_EDITION],
```
jpggvilaca: i see this operation being done a few times. What if we move the featureflag check inside the hook?

web_ui/src/pages/user-management/users/workspace-users/actions/workspace-user-actions.component.tsx
```diff
@@ -77,7 +79,10 @@ export const WorkspaceUserActions = ({ activeUser, user, users }: UserActionsPro
 
     //Note: specialCondition is used to allow user edit/delete him/herself
     return (
-        <HasPermission operations={[OPERATION.MANAGE_USER]} specialCondition={isOwnAccount}>
+        <HasPermission
+            operations={FEATURE_FLAG_WORKSPACE_ACTIONS ? [OPERATION_NEW.MANAGE_USER] : [OPERATION_OLD.MANAGE_USER]}
```
jpggvilaca: same comment for this check being done multiple times. We can add a `getManageUserOperation ` method that makes this check. 

e.g. ```const getManageUserOperation = () => {
    return FEATURE_FLAG_WORKSPACE_ACTIONS ? OPERATION_NEW.MANAGE_USER : OPERATION_OLD.MANAGE_USER;
};```

then once this "migration" is done you can replace the method, or hardcode the return value on the method.

web_ui/src/pages/project-details/components/project-sidebar/project-sidebar.component.tsx
pplaskie: It's not always that OPERATION_NEW.STH matching OPERATION_OLD.STH - sometimes operation is different in new way. So I believe will be more complications than just leaving it.

web_ui/src/pages/user-management/users/workspace-users/actions/workspace-user-actions.component.tsx
pplaskie: I did this way because when we would like to remove this FF we would just leave proper operation in this particular place, without removing function and then copying result.  But I can change if you insist ;)

web_ui/src/pages/user-management/users/users.component.tsx
```diff
@@ -5,7 +5,8 @@ import { ComponentProps, useState } from 'react';
 
 import { useUsers } from '@geti/core/src/users/hook/use-users.hook';
 import { RESOURCE_TYPE, User, UsersQueryParams } from '@geti/core/src/users/users.interface';
-import { Flex } from '@geti/ui';
+import { Button, Flex } from '@geti/ui';
```
pplaskie: commited by mistake - should not be included to this PR, I'll revert

web_ui/src/pages/project-details/components/project-sidebar/project-sidebar.component.tsx
pplaskie: Merged both operations enums into one, inlcuding new keys. No need to use FF in this place.

web_ui/src/pages/user-management/users/workspace-users/actions/workspace-user-actions.component.tsx
pplaskie: Merged both operations enums into one, inlcuding new keys. No need to use FF in this place.

web_ui/src/shared/components/has-permission/utils.ts
```diff
@@ -20,15 +20,47 @@ const WORKSPACE_CONTRIBUTOR_PERMISSION = {
     resourceType: RESOURCE_TYPE.WORKSPACE,
 };
 
-const PROJECT_MANAGER_PERMISSION = {
+const PROJECT_ADMIN_PERMISSION = {
```
dwesolow: Will server return `project_admin` or `project_manager`?

web_ui/src/pages/user-management/workspaces/workspaces.test.tsx
```diff
@@ -13,6 +17,18 @@ jest.mock('../../../providers/workspaces-provider/workspaces-provider.component'
     useWorkspaces: jest.fn(),
 }));
 
+jest.mock('@geti/core/src/users/hook/use-users.hook', () => ({
+    useUsers: jest.fn(() => ({
+        useGetUserQuery: jest.fn(),
+        useActiveUser: jest.fn(),
```
dwesolow, line 23: could we mock api service instead of hook?

web_ui/src/shared/components/has-permission/utils.ts
pplaskie: project_manager, I've changed it because I think that could be confusing if we have workspace admin, organization admin and project manager. At least it will be consistent at ui side, wdyt?

web_ui/src/pages/user-management/workspaces/workspaces.test.tsx
pplaskie: Why do you think it will be better?

dwesolow: I think that mocking API layer is better than mocking hook because hooks might have different business logic which we'd like to keep while testing and we just need to pass in memory service to the render function. When mocking hook we need to mock not only the return data but we miss all the other tanstack query attributes like `isPending`, `isSuccess` etc which we hide by using `@ts-expect-error ...`.

web_ui/src/shared/components/has-permission/utils.ts
dwesolow: Hmm, I don't have strong opinion but I'm more into `project_manager` because I'd not want to introduce ambiguity.

pplaskie: Ok, I'll revert

web_ui/src/pages/user-management/workspaces/workspaces.test.tsx
ActiveChooN: I don't think it's a big deal here. However, it's true that the more common way is to mock the data at the service level, since we have the `createInMemoryUsersService`.

web_ui/src/pages/user-management/workspaces/workspaces.test.tsx
```diff
@@ -23,9 +39,80 @@ describe('Workspaces', () => {
...
+                    resourceId: organizationId,
+                    role: USER_ROLE.ORGANIZATION_CONTRIBUTOR,
+                },
+            ],
+        });
+
+        // @ts-expect-error We only care about data property
+        jest.mocked(useUsers).mockReturnValue({ useActiveUser: () => ({ data: mockedContributorUser }) });
+
+        await render(<Workspaces />, { featureFlags: { FEATURE_FLAG_WORKSPACE_ACTIONS: true } });
```
ActiveChooN: np: `render` is sync func so no need in await

web_ui/src/pages/user-management/workspaces/workspaces.test.tsx
dwesolow: Imagine that while working on some feature that already exists and has unit tests that mock the hook. When mock returns only `data` but you added `isPending` in the feature implementation. That might force you to update the unit tests. The other example that we already faced while updating the tanstack query is that when they changed the API we had to update hooks in both places, features and unit tests. If we had mocked services everywhere, we would not need to unit tests. :)

web_ui/src/pages/user-management/users/workspace-users/actions/workspace-user-actions.component.tsx
jpggvilaca: I trust your judgement, my point was just to point something that could have been missed, tbh it's a bit hard for me to review this one, but lgtm : )

web_ui/src/pages/user-management/workspaces/workspaces.test.tsx
pplaskie: I'll leave my previous solution. I cannot make it work and don't want to spend more time on it right now (we have similar implementation in many place). If you really want to change it, please try by yourself whenever you have time or I will do that while continue working on the functionality.

jpggvilaca: We are not consistent in out project but we're going in the right direction (the direction that Dawid mentioned) but this PR is blocking further work so let's weight the pros and cons here and allow this change to be done on later PRs. For now we can leave it if it can't be done quickly. Let's merge it and on the next PR let's update this.

jpggvilaca (approved):

web_ui/src/pages/user-management/workspaces/workspaces.test.tsx
dwesolow: > I'll leave my previous solution. I cannot make it work and don't want to spend more time on it right now (we have similar implementation in many place). If you really want to change it, please try by yourself whenever you have time or I will do that while continue working on the functionality.

It was not a blocker comment, just recommendation/good practice. Feel free to leave as it is and get back to it later :).

dwesolow (approved):

PR #431, opened 2025-06-11
Make query and mutation meta typesafe

MarkRedeman:
## 📝 Description

This adds a module declaration so that our query client becomes aware of the meta info we use when handling errors.

## ✨ Type of Change

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/providers/query-client-provider/query-client-provider.component.tsx
+++ b/web_ui/src/providers/query-client-provider/query-client-provider.component.tsx
@@ -16,6 +16,19 @@ import { isAxiosError } from 'axios';
 import { NOTIFICATION_TYPE } from '../../notification/notification-toast/notification-type.enum';
 import { useNotification } from '../../notification/notification.component';
 
+declare module '@tanstack/react-query' {
+    interface Register {
+        queryMeta: {
+            notifyOnError?: boolean;
+            errorMessage?: string;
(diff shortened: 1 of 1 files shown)


jpggvilaca (approved):

web_ui/src/providers/query-client-provider/query-client-provider.component.tsx
```diff
@@ -16,6 +16,19 @@ import { isAxiosError } from 'axios';
 import { NOTIFICATION_TYPE } from '../../notification/notification-toast/notification-type.enum';
 import { useNotification } from '../../notification/notification.component';
 
+declare module '@tanstack/react-query' {
+    interface Register {
+        queryMeta: {
```
dwesolow, line 21: IIRC `meta` is not obligatory, should it be `queryMeta?: { ... }`?

MarkRedeman: This still makes it optional (there's a lot of other queries that don't use `meta`). If we make one of the fields here nonoptional then any query that uses a meta field is required to add those.
Since we're mostly aiming to use these as opt-in I figured it would be better to make them optional.

camiloHimura (approved):

PR #442, opened 2025-06-12
ITEP-32138 Use training configuration for subset splitting [PART 1]

dwesolow:
## 📝 Description

In this PR I use new `training_configuration` endpoint that returns parameters that can be configured for training.
This is the first part of many, will try to keep PRs as small as possible. For now we change the parameters locally, in one of the next PR I'll add a logic to update the server.

JIRA: ITEP-32138

## ✨ Type of Change

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/data-management.component.tsx
+++ b/web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/data-management.component.tsx
@@ -4,30 +4,35 @@
 import { FC } from 'react';
 
 import { View } from '@geti/ui';
+import { isEmpty } from 'lodash-es';
 
-import { ConfigurableParametersTaskChain } from '../../../../../../../core/configurable-parameters/services/configurable-parameters.interface';
+import { TrainingConfiguration } from '../../../../../../../core/configurable-parameters/services/configuration.interface';
 import { DataAugmentation } from './data-augmentation/data-augmentation.component';
 import { Filters } from './filters/filters.component';
 import { Tiling } from './tiling/tiling.component';
 import { TrainingSubsets } from './training-subsets/training-subsets.component';
 
 interface DataManagementProps {
-    configParameters: ConfigurableParametersTaskChain;
+    trainingConfiguration: TrainingConfiguration;
 }
 
-const getTilingParameters = (configParameters: ConfigurableParametersTaskChain) => {
-    return configParameters.components.find((component) => component.header === 'Tiling');
+const getTilingParameters = (_configParameters: TrainingConfiguration) => {
+    return undefined;
 };
 
-export const DataManagement: FC<DataManagementProps> = ({ configParameters }) => {
-    const tilingParameters = getTilingParameters(configParameters);
+export const DataManagement: FC<DataManagementProps> = ({
+    trainingConfiguration,
+}) => {
+    const tilingParameters = getTilingParameters(trainingConfiguration);
 
     return (
         <View>
             {/* Not supported in v1 of training flow revamp <BalanceLabelsDistribution /> */}
-            <TrainingSubsets />
+            <TrainingSubsets
+                subsetsConfiguration={trainingConfiguration.datasetPreparation.subsetSplit}
+            />
             {tilingParameters !== undefined && <Tiling tilingParameters={tilingParameters} />}
-            <DataAugmentation />
+            {!isEmpty(trainingConfiguration.datasetPreparation.augmentation) && <DataAugmentation />}
             <Filters />
             {/* Not supported in v1 of training flow revamp <RemovingDuplicates /> */}
         </View>
--- a/web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/data-management.component.tsx
+++ b/web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/data-management.component.tsx
@@ -14,6 +14,9 @@ import { TrainingSubsets } from './training-subsets/training-subsets.component';
 
 interface DataManagementProps {
     trainingConfiguration: TrainingConfiguration;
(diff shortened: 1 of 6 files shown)


web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/data-management.component.tsx
```diff
@@ -4,30 +4,37 @@
...
+    trainingConfiguration: TrainingConfiguration;
+    onUpdateTrainingConfiguration: (
+        updateFunction: (config: TrainingConfiguration | undefined) => TrainingConfiguration | undefined
+    ) => void;
 }
 
-const getTilingParameters = (configParameters: ConfigurableParametersTaskChain) => {
-    return configParameters.components.find((component) => component.header === 'Tiling');
+const getTilingParameters = (_configParameters: TrainingConfiguration) => {
+    return undefined;
```
dwesolow, line 23: Will be updated later once supported by backend

web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/training-subsets/training-subsets.component.tsx
```diff
@@ -121,26 +125,105 @@ const SubsetsDistribution: FC<SubsetsDistributionProps> = ({
...
+                [TRAINING_SUBSET_KEY]: trainingSubsetValue,
+                [VALIDATION_SUBSET_KEY]: validationSubsetValue,
+                [TEST_SUBSET_KEY]: testSubsetValue,
+            };
+
+            newConfig.datasetPreparation.subsetSplit = config.datasetPreparation.subsetSplit.map((parameter) => {
+                if (
+                    TRAINING_SUBSET_KEY === parameter.key ||
+                    VALIDATION_SUBSET_KEY === parameter.key ||
+                    TEST_SUBSET_KEY === parameter.key
```
dwesolow, line 188: The trick with `[...].includes(parameter.key)` didn't work here because ts didn't recognize that this key is part of `KEY_VALUE_MAP` and complained that `key` is type of `any`.

web_ui/src/pages/project-details/components/project-models/train-model-dialog/use-train-model-state.hook.ts
```diff
@@ -40,6 +43,33 @@ const getActiveModelTemplateId = (
...
+    projectIdentifier: ProjectIdentifier;
+    selectedTaskId: string;
+    selectedModelTemplateId: string | null;
+}) => {
+    const { data } = useTrainingConfigurationQuery(projectIdentifier, {
+        modelManifestId: selectedModelTemplateId,
+        taskId: selectedTaskId,
+    });
+
+    const [trainingConfiguration, setTrainingConfiguration] = useState<TrainingConfiguration | undefined>(data);
```
dwesolow, line 61: The reason for this local state is that we want to update the parameters locally first and update them on the server only when user starts a training.

jpggvilaca (dismissed):

web_ui/src/pages/project-details/components/project-models/train-model-dialog/use-train-model-state.hook.ts
```diff
@@ -40,6 +43,33 @@ const getActiveModelTemplateId = (
...
+
+    useEffect(() => {
```
jpggvilaca, line 63: why do we need this useEffect? if we're exposing the setState callback then the consumer will be responsible for updating this. If the consumer is not doing anything, then we need to expose it?

dwesolow: `trainingConfiguration` is a local state that is used for updating config parameters and it will be used (after some transformation) as a payload while sending a training request.
The need for `useEffect` is that when `useQuery` subscribes and sends request, `data` is `undefined` and once query is finished then `data` gets the real data and is no longer `undefined`. That's why I cannot just do `useState(data)` because it would never be value other than `undefined`.

web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/training-subsets/training-subsets.component.tsx
pplaskie: What about creating enum with this values 137-139?

dwesolow: What would be benefit of such change?

web_ui/src/pages/project-details/components/project-models/train-model-dialog/use-train-model-state.hook.ts
jpggvilaca: Could we not then use the query's `onSuccess`?

dwesolow: `onSuccess` was removed in react query v5 (only available in mutations) 😋

jpggvilaca: ah you're right! completely forgot :D ty

jpggvilaca (approved):

web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/training-subsets/training-subsets.component.tsx
pplaskie: It feels nicer and then you will be able to use include but with  (parameter.key as <enumName>) instead of this tree lines

pplaskie (approved):

PR #452, opened 2025-06-13
ITEP-66305 - Smart tools package [Part 3]

jpggvilaca:
## 📝 Description

* Add RITM tool to smart-tools package (Move RITM class and onnx files)
* Update interfaces and consumers

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/packages/smart-tools/src/ritm/ritm.ts
```diff
@@ -0,0 +1,339 @@
...
+import { sessionParams } from '../utils/wasm-utils';
+import { MainModelResponse, Models, RITMContour, RITMPoint, TEMPLATE_SIZE } from './interfaces';
+
+export class RITM {
+    private CV: OpenCVTypes.cv;
+    models: Models | undefined;
+    image: OpenCVTypes.Mat | undefined;
+    mask: OpenCVTypes.Mat | undefined;
+
+    static async loadRITM(cvInstance: OpenCVTypes.cv, imageData: ImageData) {
```
jpggvilaca: added this static method so we can load everything necessary for the algorithm to run without explicitly calling it like before. The consumer shouldnt care how/when things are run. The reason for it being static is because we cant run async code inside a constructor.

web_ui/packages/smart-tools/src/ritm/ritm.ts
```diff
@@ -0,0 +1,346 @@
...
+import * as ort from 'onnxruntime-web';
+import type OpenCVTypes from 'OpenCVTypes';
+
+import { Point, RegionOfInterest, Shape, ShapeType } from '../shared/interfaces';
+import { approximateShape, concatFloat32Arrays, isPolygonValid, loadSource, stackPlanes } from '../utils/utils';
+import { sessionParams } from '../utils/wasm-utils';
+import { MainModelResponse, Models, RITMContour, RITMPoint, TEMPLATE_SIZE } from './interfaces';
+
+export class RITM {
+    private CV: OpenCVTypes.cv;
```
jpggvilaca, line 15: copy paste of RITM class (only added opencv on this line and the method below on line 20)

web_ui/packages/smart-tools/src/utils/utils.ts
```diff
@@ -36,3 +36,71 @@ export const approximateShape = (CV: OpenCVTypes.cv, contour: OpenCVTypes.Mat, i
 
     return newContour;
 };
+
+export const concatFloat32Arrays = (arrays: Float32Array[]) => {
```
dwesolow, line 40: Idea: if the util is only used by tool X, I think it would be better to move it to `X/utils.ts` file so keep them separate. I'd be easier to maintain/refactor later just in case.

web_ui/src/pages/annotator/hooks/use-interactive-segmentation.hook.ts
```diff
@@ -30,48 +29,35 @@ export const useInteractiveSegmentation = ({
...
             }
+
+            setIsDrawing(false);
         };
-    }, [worker]);
 
-    useEffect(() => {
-        return () => setIsDrawing(false);
-    }, [setIsDrawing]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
dwesolow: Is the reason to add this comment only `setIsDrawing`? If yes we can safely add it as dependency because it has stable reference throughout the whole lifecycle.

web_ui/packages/smart-tools/src/ritm/ritm.ts
dwesolow: If this is the only way to create and use rtim it would be useful to make constructor private so consumer would not be allowed to do `new RITM(imageData)` (which might save some time for them while debugging 😄 )

web_ui/packages/smart-tools/src/utils/utils.ts
jpggvilaca: Absolutely. But for now you only see it being used once because we only have 1 tool. After all of them we'll do this bookeeping, or else i will move things back and forth

web_ui/packages/smart-tools/src/ritm/ritm.ts
jpggvilaca: Indeed, we can! Ill update it

web_ui/src/pages/annotator/hooks/use-interactive-segmentation.hook.ts
jpggvilaca: indeed but i'd rather be very explicit that i dont want any dependencies here. No strong opinion though

dwesolow: I just don't like adding this comment because it might get out of control (e.g. adding dependency which should call `useEffect` but actually it would not, aka hard to debug bug)

web_ui/packages/smart-tools/src/utils/utils.ts
dwesolow: TBH I didn't check the usage, just an idea ;)

jpggvilaca: i know :D hence my comment. I agree and did think about it. Just not now to avoid double work

web_ui/src/pages/annotator/hooks/use-interactive-segmentation.hook.ts
jpggvilaca: I still prefer the comment, but i reverted since it's also not related with my PR

web_ui/packages/smart-tools/src/ritm/ritm.ts
jpggvilaca: Updated

web_ui/src/webworkers/ritm.worker.ts
```diff
@@ -1,372 +1,38 @@
...
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-// Dependencies get bundled into the worker
-
-import { OpenCVLoader } from '@geti/smart-tools';
-import { expose } from 'comlink';
-import ndarray from 'ndarray';
-import ops from 'ndarray-ops';
-import * as ort from 'onnxruntime-web';
+import { OpenCVLoader, RITM } from '@geti/smart-tools';
```
dwesolow, line 4: when do u plan adding `@geti/smat-tools/{tool-name}` imports?

jpggvilaca: Not sure. But probably only after im done with all of them.

dwesolow (dismissed):

web_ui/src/pages/annotator/hooks/use-interactive-segmentation.hook.ts
```diff
@@ -103,22 +87,10 @@ export const useInteractiveSegmentation = ({
         wsInstance?.current?.reset();
     };
 
-    const loadImage = (imageData: ImageData) => {
```
MarkRedeman: Can you clarify why this was removed?

web_ui/src/pages/annotator/tools/ritm-tool/ritm-tool.component.tsx
```diff
@@ -142,7 +134,7 @@ export const RITMTool = ({ annotationToolContext }: ToolAnnotationContextProps):
         const newPoint = { x, y, positive };
         const points = result ? [...result.points, newPoint] : [newPoint];
 
-        execute(clampBox(currentBox, ROI), points, outputShape);
+        execute(image, clampBox(currentBox, ROI), points, outputShape);
```
MarkRedeman: Wouldn't this load the same image every time we execute RITM? I feel like this is something we should avoid.

jpggvilaca: It will, nice catch : ) . I reverted all refactors to keep them separate from this PR. Interface and loadImage restored. Please check the latest commit

web_ui/src/pages/annotator/hooks/use-interactive-segmentation.hook.ts
jpggvilaca: Replied below on the other comment. Changes reverted

dwesolow (approved):

web_ui/packages/smart-tools/src/utils/utils.ts
```diff
@@ -36,3 +36,71 @@ export const approximateShape = (CV: OpenCVTypes.cv, contour: OpenCVTypes.Mat, i
...
+        });
+
+        return concatFloat32Arrays(stackedPlanes.map((m) => m.data32F));
+    } finally {
+        stackedPlanes.map((p) => p.delete());
+        matPlanes?.delete();
+    }
+};
+
+export const isPolygonValid = (polygon: Polygon | null): boolean => {
```
camiloHimura, line 90: In our current implementation, Smart Tools and the Polygon tool share the same utility. This new implementation, however, is specific to Smart Tools and uses different code. Would it be possible to consolidate and just keep one utility?

jpggvilaca: The reason for this is that the original `isPolygonValid` uses `clipper-js`. I did not want to couple the 3rd party lib to our package. The second reason was that im moving 1 tool at a time. Perhaps once we move them all we will also move `clipper-js`, but im not sure. What do you guys think?

camiloHimura: Got it, it’s not a blocker right now; I just wanted to understand the reasoning behind it. But since you already have a utility that doesn’t require Clipper, I’d vote to use this one for all cases, including the Polygon tool.

camiloHimura (approved):

web_ui/packages/smart-tools/src/utils/utils.ts
MarkRedeman: I'd prefer to keep the old version. This new version doesn't really seem correct to me.
For long term I think we should replace the clipper implementation by opencv.

web_ui/packages/smart-tools/src/utils/utils.ts
```diff
@@ -36,3 +36,71 @@ export const approximateShape = (CV: OpenCVTypes.cv, contour: OpenCVTypes.Mat, i
...
+
+    const points = polygon.points;
+    if (points.length < 3) return false;
+
+    let area = 0;
+    for (let i = 0, j = points.length - 1; i < points.length; j = i++) {
+        area += (points[j].x + points[i].x) * (points[j].y - points[i].y);
+    }
+
+    area = Math.abs(area / 2);
```
MarkRedeman, line 103: This looks wrong to me: we aren't subtracting overlapping areas here, so you could draw a very thin angled rectangle that would have a small area but this function would return a very large area.
I'd prefer if we keep the old implementation.

web_ui/packages/smart-tools/src/utils/utils.ts
jpggvilaca: That is also fine. I'll just need to copy all necessary utils and add clipper-js as a dependency (at least temporarily)

jpggvilaca:
Too many changes made on the side. Will start from scratch on another PR

PR #472, opened 2025-06-18
ITEP-32138 Integrate augmentation from new training configuration [PART 4]

dwesolow:
## 📝 Description

This PR adds augmentation parameters to the advanced training dialog. I also refactored a bit how we display dependent parameters (by dependent parameters I mean that we have 1 attribute that has `key: "enable", type: "bool"` and is followed by X other parameters - in that case if first attribute is set to off then other parameters should be disabled because they don't make any impact on the training. I added small indentation so user can see that those parameters are dependent).

image

Env to test: https://10.55.252.21.
You need to use this branch locally + override `training_configuration` network request. In `dataset_prepratation` section paste
```json
"augmentation": {
            "center_crop": [
              {
                "default_value": false,
                "description": "Whether to apply center cropping to the image",
                "key": "enable",
                "name": "Enable center crop",
                "type": "bool",
                "value": true
              },
              {
                "default_value": 1.0,
                "description": "Ratio of original dimensions to keep when cropping",
                "key": "ratio",
                "max_value": null,
                "min_value": 0.0,
                "name": "Crop ratio",
                "type": "float",
                "value": 0.6
              }
            ],
            "random_affine": [
              {
                "default_value": false,
                "description": "Whether to apply random affine transformations to the image",
                "key": "enable",
                "name": "Enable random affine",
                "type": "bool",
                "value": true
              },
              {
                "default_value": 0.0,
                "description": "Maximum rotation angle in degrees",
                "key": "degrees",
                "max_value": null,
                "min_value": 0.0,
                "name": "Rotation degrees",
                "type": "float",
                "value": 15.0
              },
              {
                "default_value": 0.0,
                "description": "Maximum horizontal translation as a fraction of image width",
                "key": "translate_x",
                "max_value": null,
                "min_value": null,
                "name": "Horizontal translation",
                "type": "float",
                "value": 0.0
              },
              {
                "default_value": 0.0,
                "description": "Maximum vertical translation as a fraction of image height",
                "key": "translate_y",
                "max_value": null,
                "min_value": null,
                "name": "Vertical translation",
                "type": "float",
                "value": 0.0
              },
              {
                "default_value": 1.0,
                "description": "Scaling factor for the image during affine transformation",
                "key": "scale",
                "max_value": null,
                "min_value": null,
                "name": "Scale factor",
                "type": "float",
                "value": 1.0
              }
            ],
            "tiling": [
              {
                "default_value": false,
                "description": "Whether to apply tiling to the image",
                "key": "enable",
                "name": "Enable tiling",
                "type": "bool",
                "value": true
              },
              {
                "default_value": false,
                "description": "Whether to use adaptive tiling based on image content",
                "key": "adaptive_tiling",
                "name": "Adaptive tiling",
                "type": "bool",
                "value": false
              },
              {
                "default_value": 128,
                "description": "Size of each tile in pixels",
                "key": "tile_size",
                "max_value": null,
                "min_value": 0,
                "name": "Tile size",
                "type": "int",
                "value": 256
              },
              {
                "default_value": 64,
                "description": "Overlap between adjacent tiles in pixels",
                "key": "tile_overlap",
                "max_value": null,
                "min_value": 0,
                "name": "Tile overlap",
                "type": "int",
                "value": 64
              }
            ]
        },
```

JIRA: ITEP-32138

## ✨ Type of Change

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


jpggvilaca (approved):

web_ui/src/pages/project-details/components/project-models/train-model-dialog/advanced-settings/data-management/data-management.component.tsx
```diff
@@ -19,12 +19,21 @@ interface DataManagementProps {
     ) => void;
 }
 
+const getAugmentationParameters = (configuration: TrainingConfiguration) => {
+    const augmentation = structuredClone(configuration.datasetPreparation.augmentation);
+
+    delete augmentation['tiling'];
```
camiloHimura, line 25: I’m not a fan of using `delete` since it mutates the original object, which can lead to hard-to-track bugs. It’s probably not an issue in this case, but I still think it’s something we should avoid when possible.

dwesolow: It mutates original object that's why I do it on a deep copy of the object :)

camiloHimura: got it 👍🏼

camiloHimura (approved):

PR #481, opened 2025-06-18
Save last visited workspace and open it as a default

pplaskie:
## 📝 Description

Last visited workspace is saved in the localStorage and open as a default. If no saved workspace then first one is opened as a default.

I'll add tests afterwards, if the code will be accepted ;)

JIRA: ITEP-69585

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/landing-page/workspaces-tabs/workspaces-tabs.component.tsx
+++ b/web_ui/src/pages/landing-page/workspaces-tabs/workspaces-tabs.component.tsx
@@ -1,7 +1,7 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-import { Key, useEffect } from 'react';
+import { Key } from 'react';
 
 import { paths } from '@geti/core';
 import { useFeatureFlags } from '@geti/core/src/feature-flags/hooks/use-feature-flags.hook';
@@ -19,7 +19,6 @@ import {
 } from '@geti/ui';
 import { Add } from '@geti/ui/icons';
 import { useNavigate } from 'react-router-dom';
-import { useLocalStorage } from 'usehooks-ts';
 
 import { useOrganizationIdentifier } from '../../../hooks/use-organization-identifier/use-organization-identifier.hook';
 import { usePinnedCollapsedItems } from '../../../hooks/use-pinned-collapsed-items/use-pinned-collapsed-items.hook';
@@ -33,7 +32,6 @@ import { EditNameDialog } from '../../../shared/components/edit-name-dialog/edit
 import { HasPermission } from '../../../shared/components/has-permission/has-permission.component';
 import { OPERATION } from '../../../shared/components/has-permission/has-permission.interface';
 import { TabItem } from '../../../shared/components/tabs/tabs.interface';
-import { getLastOpenedWorkspaceKey, LOCAL_STORAGE_KEYS } from '../../../shared/local-storage-keys';
 import { getUniqueNameFromArray, hasEqualId } from '../../../shared/utils';
 import { MAX_LENGTH_OF_WORKSPACE_NAME, MIN_LENGTH_OF_WORKSPACE_NAME } from '../../user-management/workspaces/utils';
 import { LandingPageWorkspace as Workspace } from '../landing-page-workspace/landing-page-workspace.component';
@@ -46,8 +44,8 @@ const MAX_NUMBER_OF_DISPLAYED_WORKSPACES = 6;
 const usePinnedCollapsedWorkspaces = () => {
     const navigate = useNavigate();
     const { organizationId } = useOrganizationIdentifier();
-
     const { workspaces, workspaceId: selectedWorkspaceId } = useWorkspaces();
+
     const [pinnedWorkspaces, collapsedWorkspaces, dispatch] = usePinnedCollapsedItems(
         workspaces,
         selectedWorkspaceId,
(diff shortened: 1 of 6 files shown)


web_ui/src/pages/landing-page/workspaces-tabs/workspaces-tabs.component.tsx
```diff
@@ -113,6 +117,10 @@ export const WorkspacesTabs = (): JSX.Element => {
         children: <Workspace />,
     }));
 
+    useEffect(() => {
+        setLastWorkspaceId(selectedWorkspaceId);
+    }, [selectedWorkspaceId]);
```
dwesolow: It's not recommended to use `useEffect` in this case since we can update `lastWorkspaceId` in the handler itself instead of in reaction of the `selectedWorkspaceId` change. In this case we could move setting `lastWorkspaceId` logic into `usePinnedCollapsedWorkspaces` and do:
```
const selectWorkspace = (workspaceId: string) => {
        navigate(paths.workspace({ organizationId, workspaceId }));
        setLastWorkspaceId(workspaceId);
};
```

Then we are always sure that `selectedWorkspaceId` is always in sync with `lastWorkspaceId` (whenever we'd use `usePinnedCollapsedWorkspaces`)

web_ui/src/pages/landing-page/workspaces-tabs/workspaces-tabs.component.tsx
```diff
@@ -32,6 +33,7 @@ import { EditNameDialog } from '../../../shared/components/edit-name-dialog/edit
 import { HasPermission } from '../../../shared/components/has-permission/has-permission.component';
 import { OPERATION } from '../../../shared/components/has-permission/has-permission.interface';
 import { TabItem } from '../../../shared/components/tabs/tabs.interface';
+import { getLastOpenedWorkspaceKey, LOCAL_STORAGE_KEYS } from '../../../shared/local-storage-keys';
```
dwesolow: WDYT of extracting a custom hook called `useLastWorkspaceId` and use it here and in `redirect.component.tsx`? I think it could be nice solution (it adds small abstraction and hides a logic that we use local storage for this).

web_ui/src/pages/landing-page/workspaces-tabs/workspaces-tabs.component.tsx
pplaskie: And what about the case when user open url from different workspace? We would also need to save it but selectWorkspace function is now called

dwesolow: It's already handled by `useFirstWorkspaceIdentifier` in `useWorkspaces` (this hook is used by `usePinnedCollapsedWorkspaces`)

pplaskie: i mean that I cannot call setLastWorkspaceId only in selectWorkspace because also when user opens url it should be saved

dwesolow: Ahm, good point, we need to update if on first mount. Then I think we need to use `useEffect` 😞

web_ui/src/providers/workspaces-provider/use-workspace-identifier.hook.ts
```diff
@@ -14,5 +16,14 @@ export const useWorkspaceIdentifier = (): WorkspaceIdentifier => {
...
     const { workspaceId = '' } = useParams<Pick<WorkspaceIdentifier, 'workspaceId'>>();
 
+    const { lastWorkspaceId, setLastWorkspaceId } = useLastWorkspace(organizationId, workspaceId);
+
+    useEffect(() => {
+        if (!isEqual(lastWorkspaceId, workspaceId)) {
+            setLastWorkspaceId(workspaceId);
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [workspaceId]);
```
dwesolow: **Suggestion**: could we move this part to `useLastWorkspace` to make that hook responsible for handling it? The benefit of it would be that we don't need to expose `setLastWorkspaceId` from the hook.

web_ui/src/providers/workspaces-provider/use-workspace-identifier.hook.ts
```diff
@@ -14,5 +16,14 @@ export const useWorkspaceIdentifier = (): WorkspaceIdentifier => {
     const { organizationId } = useOrganizationIdentifier();
     const { workspaceId = '' } = useParams<Pick<WorkspaceIdentifier, 'workspaceId'>>();
 
+    const { lastWorkspaceId, setLastWorkspaceId } = useLastWorkspace(organizationId, workspaceId);
+
+    useEffect(() => {
+        if (!isEqual(lastWorkspaceId, workspaceId)) {
+            setLastWorkspaceId(workspaceId);
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
dwesolow: **Suggestion:** I think we don't need this comment. When `lastWorkspaceId` is different than `workspaceId` then we want to update it, if no then `setLastWorkspaceId` is not invoked. `setLastWorkspaceId` has stable reference so it never cause `useEffect` to trigger.

dwesolow: Discussed offline: we decided to leave as it is because in `redirect.component.tsx` we don't want to set `lastWorkspaceId` to the latest workspace.

dwesolow (approved):

web_ui/src/providers/workspaces-provider/use-workspace-identifier.hook.ts
dwesolow: Discussed offline, this change would break in `redirect.component.tsx` because it would set always the first workspace.

jpggvilaca (approved):

PR #482, opened 2025-06-18
:chore: Remove notification dependency from use-personal-token hook

jpggvilaca:
## 📝 Description

* One less thing to worry about when moving PAT to @geti/core
* The hook should be responsible for the queries, not for extra actions.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/user-management/personal-access-token-page/components/personal-access-tokens-menu.component.tsx
+++ b/web_ui/src/pages/user-management/personal-access-token-page/components/personal-access-tokens-menu.component.tsx
@@ -11,7 +11,9 @@ import { AxiosError } from 'axios';
 import dayjs from 'dayjs';
 
 import {
+    DELETE_ERROR,
     DELETE_MESSAGE,
+    UPDATE_ERROR,
     UPDATE_MESSAGE,
     usePersonalAccessToken,
 } from '../../../../core/personal-access-tokens/hooks/use-personal-access-token.hook';
@@ -68,7 +70,7 @@ export const PersonalAccessTokenMenu = ({ token }: PersonalAccessTokenMenuProps)
                         addNotification({ message: UPDATE_MESSAGE, type: NOTIFICATION_TYPE.DEFAULT });
                     },
                     onError: (error: AxiosError) => {
-                        const message = error?.message ?? DELETE_MESSAGE;
+                        const message = error?.message ?? UPDATE_ERROR;
                         addNotification({ message, type: NOTIFICATION_TYPE.ERROR });
                     },
                 }
@@ -81,8 +83,11 @@ export const PersonalAccessTokenMenu = ({ token }: PersonalAccessTokenMenuProps)
             deletePersonalAccessTokenMutation.mutate(
                 { organizationId, userId: activeUser.id, tokenId: token.id },
                 {
+                    onSuccess: () => {
+                        addNotification({ message: DELETE_MESSAGE, type: NOTIFICATION_TYPE.DEFAULT });
+                    },
                     onError: (error: AxiosError) => {
-                        const message = error?.message ?? DELETE_MESSAGE;
+                        const message = error?.message ?? DELETE_ERROR;
                         addNotification({ message, type: NOTIFICATION_TYPE.ERROR });
                     },
                 }
--- a/web_ui/src/pages/user-management/personal-access-token-page/components/personal-access-tokens-menu.component.tsx
+++ b/web_ui/src/pages/user-management/personal-access-token-page/components/personal-access-tokens-menu.component.tsx
@@ -7,11 +7,18 @@ import { useUsers } from '@geti/core/src/users/hook/use-users.hook';
 import { ActionButton, Flex, Item, Menu, MenuTrigger, Text } from '@geti/ui';
 import { Delete, Edit, MoreMenu } from '@geti/ui/icons';
 import { useOverlayTriggerState } from '@react-stately/overlays';
+import { AxiosError } from 'axios';
 import dayjs from 'dayjs';
 
-import { usePersonalAccessToken } from '../../../../core/personal-access-tokens/hooks/use-personal-access-token.hook';
+import {
+    DELETE_MESSAGE,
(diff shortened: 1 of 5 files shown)


dwesolow:
TBH I don't agree with this change. I like the previous implementation where the hook is responsible for the whole PAT logic/behavior. In this PR, whenever I'd like to reuse the hook, I'd need to provide `onSuccess` etc with notifications each time instead of reusing the hook with encapsulated behavior.

jpggvilaca:
> TBH I don't agree with this change. I like the previous implementation where the hook is responsible for the whole PAT logic/behavior. In this PR, whenever I'd like to reuse the hook, I'd need to provide `onSuccess` etc with notifications each time instead of reusing the hook with encapsulated behavior.

While I understand your argument, i'd disagree because the hook is doing more than 1 thing. It's handling queries AND triggering notifications. This way we keep things separate and let the UI component customise its own notification. For instance if you need a custom notification message, you would need to pass it to the query, increasing coupling.

There's also a middleground, like adding these props to the meta key, i.e.

`useMutation(trainModel, {
      meta: {
           successMessage: 'bla bla',
           errorMessage: 'bla bla'
      }
}`

But againt this is polluting the mutation itself. Furthermore. We might not even want a notification, and in that case, we cannot have this behavior because the hook doesnt allow us to.

In any case, i think we'll end up removing  MOST of them even from the components, by adding the notification trigger on the query/mutation cache itself, like we have right now on "web_ui/src/providers/query-client-provider/query-client-provider.component.tsx".

Let's see what others think as well.

camiloHimura:
That’s a good point, @jpggvilaca, but at the same time, I agree with @dwesolow that this approach would mean writing more code in every use case. Since all `onErrors` currently follow the same pattern, I think the hook should handle it internally.

jpggvilaca:
> That’s a good point, @jpggvilaca, but at the same time, I agree with @dwesolow that this approach would mean writing more code in every use case. Since all `onErrors` currently follow the same pattern, I think the hook should handle it internally.

From my POV i dont think it's more code. We just move it from the hook to the consumer UI. We often want customised notifications so in this case we dont need to pass them on to the hook but use them directly.

I'm currently working on 

```
const mutationCache = new MutationCache({
        onError: (error) => {
            // TODO: If we decide to add an "opt-out" mechanism we can use the same logic
            // as query cache and use the `meta` field
            if (isAxiosError(error)) {
                const message = error.message;
                notify.current({
                    message: typeof message === 'string' ? message : getErrorMessage(error),
                    type: NOTIFICATION_TYPE.ERROR,
                });
            }
        },
    });
```

Which follows the same approach as our queryCache. This means that you can do a cmd+shift+f for 

`addNotification({ message: getErrorMessage(error), type: NOTIFICATION_TYPE.ERROR });`

And we can remove all of those. Meaning this PR would be a non-issue once that is in place.

But if everyone's against ofc i can easily close this PR

web_ui/src/pages/user-management/personal-access-token-page/components/personal-access-tokens-menu.component.tsx
```diff
@@ -7,11 +7,20 @@ import { useUsers } from '@geti/core/src/users/hook/use-users.hook';
...
 import dayjs from 'dayjs';
 
-import { usePersonalAccessToken } from '../../../../core/personal-access-tokens/hooks/use-personal-access-token.hook';
+import {
+    DELETE_ERROR,
+    DELETE_MESSAGE,
+    UPDATE_ERROR,
+    UPDATE_MESSAGE,
+    usePersonalAccessToken,
+} from '../../../../core/personal-access-tokens/hooks/use-personal-access-token.hook';
```
MarkRedeman: I think this part feels counter intuitive to me here: on the one hand we want the consumer of the hook to choose what notifications to choose, but we end up with adding back tight coupling because we only use data from the hook's file.

camiloHimura:
> From my POV i dont think it's more code. We just move it from the hook to the consumer UI. We often want customised notifications so in this case we dont need to pass them on to the hook but use them directly.

Thanks for the context, I like that approach 👍🏼

MarkRedeman (commented):
I think that I'd prefer if we either:
1. Remove as much occurances of `useNotification` as possible and rely on inline error messages instead. So in this example instead of showing the notification at the bottom of the screen, we should give the "create" button a visual that indicates that an error happened (with the option to retry) and add a error message inside of the modal, not at place that is not related to the modal.
2. If we want to be lazy: apply the `meta` approach that we did with our queries so that we don't need to do `const { addNotification } = useNotification` everywhere.

web_ui/src/pages/user-management/personal-access-token-page/components/personal-access-tokens-menu.component.tsx
jpggvilaca: The component was already coupled with the hook, it wont work without it. Maybe it's because you see all message imports? I guess i could also move them here to this file, where they are used, wdyt?

MarkRedeman: Yeah but before the notification messages and sending the notifications were both in the `usePersonalAccessToken` hook, and now we have the messages in the hook, but sending them is being done by the consumer.
This feels odd because it gives the impression that the hook intents for notifications to be send but doesn't do so.

> I guess i could also move them here to this file, where they are used, wdyt?

Yeah that would be better imho.

But I'd prefer us to implement the first point I mentioned above (not using notifications for errors and inline them instead), I think that will result in a better UX.

jpggvilaca: Moved constants to the dialog

pplaskie (approved):

camiloHimura (approved):

dwesolow (approved):

PR #489, opened 2025-06-19
ITEP-68000 - Import keypoint detection - p2

camiloHimura:
## 📝 Description
This PR displays the project template to the user before starting the import process. The template is read-only, and a new component, `ReadonlyTemplateManager`, was added to handle this functionality.

https://github.com/user-attachments/assets/48f15143-dbca-4c1e-953c-ecd78405aada

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/create-project/components/pose-template/canvas/canvas-template.component.tsx
+++ b/web_ui/src/pages/create-project/components/pose-template/canvas/canvas-template.component.tsx
@@ -16,15 +16,13 @@ import { KeyMap } from '../../../../../shared/keyboard-events/keyboard.interface
 import { getIds, hasDifferentId } from '../../../../../shared/utils';
 import { ClosestKeypoint } from '../../../../annotator/tools/edit-tool/edit-keypoint/closest-keypoint.component';
 import { useZoom } from '../../../../annotator/zoom/zoom-provider.component';
-import { EdgeLine, getPointInRoi } from '../../../../utils';
+import { EdgeLine, getPointInRoi, TemplateState, TemplateStateWithHistory } from '../../../../utils';
 import {
     getDefaultLabelStructure,
     isDifferentLabel,
     isEdgeConnectingPoints,
     isEqualLabel,
     isNotMatchingEdge,
-    TemplateState,
-    TemplateStateWithHistory,
     updateWithLatestPoints,
 } from '../util';
 import { DrawingBox } from './drawing-box.component';
--- a/web_ui/src/pages/create-project/components/pose-template/canvas/canvas-template.component.tsx
+++ b/web_ui/src/pages/create-project/components/pose-template/canvas/canvas-template.component.tsx
@@ -16,9 +16,8 @@ import { KeyMap } from '../../../../../shared/keyboard-events/keyboard.interface
 import { getIds, hasDifferentId } from '../../../../../shared/utils';
 import { ClosestKeypoint } from '../../../../annotator/tools/edit-tool/edit-keypoint/closest-keypoint.component';
 import { useZoom } from '../../../../annotator/zoom/zoom-provider.component';
-import { getPointInRoi } from '../../../../utils';
+import { EdgeLine, getPointInRoi } from '../../../../utils';
 import {
-    EdgeLine,
     getDefaultLabelStructure,
     isDifferentLabel,
     isEdgeConnectingPoints,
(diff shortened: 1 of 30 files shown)


web_ui/src/pages/create-project/components/pose-template/readonly-template-manager.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { CanvasTemplate } from './canvas/canvas-template.component';
+import { createRoi, resizePoints, TemplateState } from './util';
+
+export interface ReadonlyTemplateManagerProps {
+    className: string;
+    scaleFactor?: number;
+    initialNormalizedState: TemplateState;
+}
+
+export const ReadonlyTemplateManager = ({
```
jpggvilaca, line 25: would it be better to have a little duplication but making a simpler template manager without any dependencies?

jpggvilaca:
One nitpick: Can we center the template (i.e. the canvas)? The skeleton being left-aligned looked weird

web_ui/src/pages/create-project/components/pose-template/readonly-template-manager.component.tsx
camiloHimura: That was my initial goal, but after seeing the number of updated files, I decided to keep it simple for now. I’ll implement it in my upcoming PR, as I’ll need to reuse this component for importing non-GETI datasets.

jpggvilaca: So you will make a simpler/duplicated one on the next PR?

camiloHimura: Yes, a simpler version. Right now, we only need to render the template, not update it (at least for now; that could become a feature requirement later on).

camiloHimura:
> One nitpick: Can we center the template (i.e. the canvas)? The skeleton being left-aligned looked weird

Good point, I’ve updated it in my latest commit.

jpggvilaca (approved):

web_ui/src/pages/create-project/components/pose-template/canvas/canvas-template.component.tsx
```diff
@@ -168,10 +167,10 @@ export const CanvasTemplate = ({
                     border: '1px solid var(--spectrum-global-color-gray-200)',
                 }}
             >
-                {isAddPointEnabled && (
+                {isAddPointEnabled && nextPointName !== null && (
```
MarkRedeman, line 170: Do we need this condition? It looks like `nextPointName` is always a string no?

web_ui/src/pages/create-project/components/pose-template/util.ts
```diff
@@ -118,6 +104,35 @@ export const updateWithLatestPoints =
...
+    const [minY, maxY] = getMaxMinPoint(points, PointAxis.Y);
+
+    const currentWidth = maxX - minX;
+    const currentHeight = maxY - minY;
+
+    const newWidth = currentWidth * scaleFactor;
+    const newHeight = currentHeight * scaleFactor;
+
+    const scaleX = newWidth / currentWidth;
+    const scaleY = newHeight / currentHeight;
```
MarkRedeman, line 118: Do you need `newWidth` and `newHeight` here? I.e.
```
scaleX = newWidth / currentWidth = (currentWidth * scaleFactor) / currentWidth = scaleFactor
```

I guess below we do need it to add offset, but it might be simpler if we use `scaleFactor` directly instead of `scaleX` and `scaleY`?

web_ui/src/pages/utils.ts
```diff
@@ -134,3 +156,22 @@ export const projectPointOnLine = ([startPoint, endPoint]: ProjectLine, point: P
...
+
+        return { id: uuidv4(), from: toLabel, to: fromLabel } as EdgeLine;
+    });
+
+    return { edges: linkedEdged, points: positions };
+};
+
+export const getMaxMinPoint = <T extends Point>(points: T[], pointAxis: PointAxis) => {
+    const minAxisValue = Math.min(...points.map((point) => point[pointAxis]));
+    const maxAxisValue = Math.max(...points.map((point) => point[pointAxis]));
```
MarkRedeman, line 175: Possible suggestion: we could also use a reduce here and get the min and max in 1 loop and without copying `points`.
I guess in our case we don't expect much more than say 100 points so the performance (in other cases doing `Math.min(...xs)` can be a lot slower than doing a reduce or for loop) shouldn't matter but wanted to share in case you think it's useful.

MarkRedeman (approved):

web_ui/src/pages/utils.ts
camiloHimura: yes, I'll address it in my next PR

PR #496, opened 2025-06-20
ITEP-68003 - Import keypoint dataset to existing project

camiloHimura:
## 📝 Description

This PR disables the import button when the mapped labels contain duplicates. It also clears the entire mapping if it's incomplete. In such cases, the user is advised to import the dataset as a single project, since all annotations would otherwise be discarded

https://github.com/user-attachments/assets/18a89f18-6086-4788-abb1-d6cf38b41cb6

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/dataset-import-to-existing-project-dialog.component.tsx
+++ b/web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/dataset-import-to-existing-project-dialog.component.tsx
@@ -3,12 +3,10 @@
 
 import { useMemo } from 'react';
 
-import { Content, Dialog, DialogContainer, Divider, Flex, Heading, Text, View } from '@geti/ui';
-import { Alert } from '@geti/ui/icons';
+import { Content, Dialog, DialogContainer, Divider, Heading, View } from '@geti/ui';
 import { OverlayTriggerState } from '@react-stately/overlays';
 
 import { DATASET_IMPORT_STATUSES } from '../../../../../core/datasets/dataset.enum';
-import { useFeatureFlags } from '../../../../../core/feature-flags/hooks/use-feature-flags.hook';
 import { isAnomalyDomain, isKeypointDetection } from '../../../../../core/projects/domains';
 import { useDatasetImportToExistingProject } from '../../../../../providers/dataset-import-to-existing-project-provider/dataset-import-to-existing-project-provider.component';
 import { matchStatus } from '../../../../../providers/dataset-import-to-existing-project-provider/utils';
@@ -17,7 +15,8 @@ import { DatasetImportProgress } from '../../../../../shared/components/dataset-
 import { useProject } from '../../../providers/project-provider/project-provider.component';
 import { DatasetImportToExistingProjectDialogButtons } from './dataset-import-to-existing-project-dialog-buttons.component';
 import { DatasetImportToExistingProjectMapLabels } from './dataset-import-to-existing-project-map-labels.component';
-import { hasDuplicatedValues, KEYPOINT_DUPLICATED_LABELS } from './utils';
+import { KeypointErrorMessage } from './keypoint-error-message.component';
+import { hasDuplicatedValues } from './utils';
 
 interface DatasetImportToExistingProjectDialogProps {
     datasetImportDialogState: OverlayTriggerState;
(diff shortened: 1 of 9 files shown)


web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/utils.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2022-2025 Intel Corporation
+// LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
+
+import { uniq } from 'lodash-es';
+
+import { Label } from '../../../../../core/labels/label.interface';
+
+export const KEYPOINT_DUPLICATED_LABELS = '- Duplicate labels are not allow in keypoint detection projects:';
```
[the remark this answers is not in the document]
camiloHimura: updated

web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/dataset-import-to-existing-project-dialog.component.tsx
```diff
@@ -47,20 +52,32 @@ export const DatasetImportToExistingProjectDialog = ({
...
+
+    const handleDialogDismiss = (): void => {
         datasetImportDialogState.close();
         setActiveDatasetImportId(undefined);
     };
 
+    const handlePrimaryAction = () => {
+        if (!activeDatasetImport) return;
+
+        if (isKeypointMapLabels && getMissingLabels(project.labels, activeDatasetImport.labelsMap)) {
```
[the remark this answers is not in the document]
camiloHimura: updated

web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/keypoint-error-message.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const duplicatedLabels = labels.filter((label) => duplicatedValues.includes(label.id));
+
+    if (isEmpty(missingLabels) && isEmpty(duplicatedLabels)) {
+        return <></>;
+    }
+
+    return (
+        <Flex gap={'size-100'} alignContent={'center'} height={'100%'} alignItems={'center'}>
+            <Alert style={{ fill: 'var(--brand-coral-cobalt)', flex: 'none' }} />
+            <Flex direction={'column'} gap={'size-100'}>
```
jpggvilaca, line 39: should we move the `duplicatedLabels` and `missingLabels` verification to the parent? To avoid any edge case where the Alert icon would be there but without any message? Also, np: should we make it Warning (yellow) instead of red?

web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/util.test.ts
```diff
@@ -0,0 +1,100 @@
+// Copyright (C) 2022-2025 Intel Corporation
```
jpggvilaca, line 1: I really like that you always add tests for everything, but can i point out just a minor np? These util tests let's try and add them on a different PR, so it doesnt clutter the actual feature you're trying to implement ^^

jpggvilaca (approved):

web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/util.test.ts
camiloHimura: It was added here because they’re core to the feature; they indicate whether the user has missing or duplicate labels. They need to be part of this PR. :)

web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/keypoint-error-message.component.tsx
camiloHimura: I didn’t want to move it to the parent because the `DatasetImportToExistingProjectDialogButtons ` already has a well-defined responsibility. Encapsulating the error handling here feels like a better way to maintain the separation of concerns.

As for the single icon case, I’ve added an early-return validation to prevent that scenario.

And I agree about the yellow color, red feels too serious.

web_ui/src/pages/project-details/components/project-dataset/dataset-import-to-existing-project/dataset-import-to-existing-project-dialog.test.tsx
```diff
@@ -205,4 +208,72 @@ describe(DatasetImportToExistingProjectDialog, () => {
...
         expect(mockedImportDatasetJob).toHaveBeenCalledWith(mockDatasetImportItem.id);
     });
+
+    describe('keypoint detection', () => {
+        const neckLabel = getMockedLabel({ name: 'neck', id: '683d4ccfd01df152c3f65ff6' });
+        const headLabel = getMockedLabel({ name: 'neck', id: '683d4ccfd01df152c3f65ff7' });
+
+        it('displays warning for duplicated keypoint labels', async () => {
+            jest.mocked(useDatasetImportToExistingProject).mockReturnValue({
+                ...jest.requireActual(
```
MarkRedeman, line 218: For future refactor: I'd prefer if we avoid mocking providers in our test code. Ideally we should mock the source (the incoming data from services) instead. This way we prevent testing our mocks instead of testing our logic.

MarkRedeman (approved):

PR #497, opened 2025-06-20
ITEP-32067 - Add backward compatibility for configurations

maxxgx:
## 📝 Description

- Add backward compatibility for mapping legacy configurable parameters to the revamped configuration, and vice versa.
Note that not all legacy parameters are compatible and are therefore deprecated. This means that if a REST API user chooses to use the legacy configuration endpoints, the following will occur:
   - They will not be able to access or set some of the new parameters, such as augmentations.
   - Setting deprecated parameters will have no effect.

Other fixes:
- Fixed wrong value for `dataset_size` parameter.
- Add query parameters for the training_configuration PATCH endpoint (e.g., `training_configuration?task_id=...&model_manifest_id=...`). These can also be specified within the payload itself.
- Add `?task_id` query parameter for project_configuration endpoints (GET and PATCH). If ?task_id is provided, the endpoint will return the configuration for a single task, which is useful for task-chain projects. Similarly, for the PATCH method, the payload can contain only the configuration for the specified task.
- Configuration can be retrieved for legacy trained models.

Note: all changes are behind feature flag.

JIRA: ITEP-32067

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


interactive_ai/services/director/app/communication/endpoints/training_configuration_endpoints.py
```diff
@@ -31,8 +31,16 @@
 
 def get_training_configuration_from_request(
     request_json: Annotated[dict, Depends(get_request_json)],
+    task_id: Annotated[str | None, Query()] = None,
+    model_manifest_id: Annotated[str | None, Query()] = None,
 ) -> PartialTrainingConfiguration:
     """Dependency to convert REST request body to PartialTrainingConfiguration."""
+    # task_id and model_manifest_id can be present in both payload and query parameters
```
leoll2, line 38: why?

leoll2 (approved):

interactive_ai/services/director/app/communication/endpoints/training_configuration_endpoints.py
maxxgx: I think it's beneficial to have this for convenience. For instance, it allows us to copy the response from the GET endpoint and use it directly as the payload for the PATCH endpoint. If `task_id` is present in both the query parameter and the payload, the one from the payload should take priority, as it is more specific, in my opinion.

PR #498, opened 2025-06-20
build: migrate OPA policies to geti repo

mpryahin:
## 📝 Description

* Move the OPA-policy bundle into the geti repository and plug it into the new build system.
* Remove environment variables that are no longer used.
* Extend the Helm-chart build target so it can optionally include additional resources.
* Make the directory be used as a temp folder for *all* build artifacts, not just Helm charts.
* Enhance the `builder` target: when no matching builder image is found for the project type, fall back to executing the wrapped make target directly in the local environment.
* Add a Github runners cleanup action to free up disk space"

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


mpryahin:
`/interactive_ai/workflows/otx_domain/trainer/otx_v2` build fails due to `no space left on device`

deploy/charts/geti-components/chart/Chart.yaml.template
```diff
@@ -16,96 +16,100 @@ appVersion: ${TAG}
...
-    repository: "file://../../../../platform/services/weights_uploader/.build"
+    repository: "file://../../../../../platform/services/weights_uploader/.build/chart"
     condition: enabled.weights_uploader
   - name: web
     version: ${TAG}
-    repository: "file://../../../../web_ui/.build"
+    repository: "file://../../../../../web_ui/.build/chart"
     condition: enabled.web_ui
+  - name: opa-bundle
+    repository: "file://../../../../../platform/services/onboarding/.build/chart"
```
michalad1:
```suggestion
    repository: "file://../../../../../platform/services/opa_bundle/.build/chart"
```

.github/components-path-filters.yml
```diff
@@ -277,19 +281,19 @@ interactive_ai/workflows/geti_domain/train:
...
-#  - interactive_ai/workflows/otx_domain/trainer/otx_v2/tests/**
-#  - interactive_ai/workflows/otx_domain/trainer/otx_v2/download_pretrained_weights.py
-#  - interactive_ai/workflows/otx_domain/trainer/otx_v2/pyproject.toml
-#  - interactive_ai/workflows/otx_domain/trainer/otx_v2/run
-#  - libs/telemetry_tools/**
-#  - libs/kafka_tools/**
-#  - Makefile.shared-python
-#  - Makefile.shared
-#  - .github/workflows/component.yml
 # interactive_ai/workflows/otx_domain/trainer/otx_v2/gpu:
```
mramotowski: I see xpu is uncommented but gpu not are there more storage requirements for that one?

.github/workflows/component.yml
```diff
@@ -55,15 +55,16 @@ jobs:
...
-      - name: Free space on runner (delete unnecessary tools)
-        run: rm -rf /opt/hostedtoolcache
-
       - name: Checkout code
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
         with:
           persist-credentials: false
           ref: ${{ inputs.ref || '' }}
 
+      - name: Initial cleanup
```
mramotowski, line 64: If I understand correctly we are cleaning it cause there is not enough space for `interactive_ai/workflows/otx_domain/trainer/otx_v2/xpu`.  Wouldn't it be better to add if statement and check if what we are running is `interactive_ai/workflows/otx_domain/trainer/otx_v2/xpu` and if not don't do it for other components? Cause this step takes from `2X seconds` to `1,2X minute` to be executed.

.github/components-path-filters.yml
mpryahin: I will double check, but this change appeared after I rebased on main

leoll2: CC @AlbertvanHouten

.github/workflows/component.yml
leoll2: @mramotowski Do you have an example run where this step takes 1.2 minutes? It is supposed to be very quick

.github/components-path-filters.yml
AlbertvanHouten: Both should be commented out like it is in main, they will be reenabled in https://github.com/open-edge-platform/geti/pull/495. This looks like a wrongly solved merge conflict.

mpryahin: Fixed

.github/workflows/component.yml
mramotowski: https://github.com/open-edge-platform/geti/actions/runs/15819308367/job/44584678337?pr=498 However, I believe that even 22 seconds is excessive, as this step is unnecessary in other components.

mpryahin: Given that the entire workflow takes 22 minutes 28 seconds, those 22 seconds amount to roughly 1 % of the total runtime, hardly excessive. Enabling this cleanup only for certain components would just clutter the workflow definition. Let’s keep it on for all components, and if it starts to take noticeably longer we can optimise it then.

michalad1 (dismissed):

.github/workflows/component.yml
AlbertvanHouten: Which workflow is taking 22 minutes? Most of the individual components take ~3 minutes to complete. Even 22 seconds is then quite a significant amount. I agree with Maciej that it would be nice to only run this step for disk space restricted workflows.

jcchr (dismissed):

.github/workflows/component.yml
mpryahin: 22 min is the full platform build

mpryahin: @mramotowski  @AlbertvanHouten Thanks, I will create a separate PR for that after merging this one to avoid bloating its scope.

michalad1 (approved):

jcchr (approved):

mramotowski (approved):

PR #511, opened 2025-06-23
Sanitize logs to prevent log injection

AlbertvanHouten:
## 📝 Description
Sanitize log messages that could potentially be vulnerable to log injection. 

Removed logic to periodically refresh logger settings. This was only set up to change the log level, but practically never used. In https://jira.devtools.intel.com/browse/ITEP-70813 I propose to set an env var in the services deployments which if changed, causes the service to restart and then also reload the new log level eliminating the need to periodically reload the log settings. 

https://jira.devtools.intel.com/browse/ITEP-70639
https://jira.devtools.intel.com/browse/ITEP-70644
https://jira.devtools.intel.com/browse/ITEP-70643

## ✨ Type of Change

- [x] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist


leoll2 (dismissed):

leoll2 (dismissed):

libs/telemetry_tools/src/geti_logger_tools/logger_config.py
```diff
@@ -81,23 +81,34 @@ def refresh_logger(file_path, log_level=None, interval=30, loop=None, use_async=
...
     :param package_name: given name of package for the logger
     :param use_async: whether to use asyncio for scheduling refresh task (if not then threading)
     :param logging_format: optional, logging format to use instead of the default one
     :return: initialized logger
     """
+
+    class SanitizeLogFilter(logging.Filter):
+        def filter(self, record: logging.LogRecord) -> bool:
+            # Sanitize the log message to prevent log injection
+            record.msg = str(record.msg).replace("\n", "\\n").replace("\r", "\\r")
```
AlexanderBarabanov: This escape does not work, I would propose to use space 
`record.msg = str(record.msg).replace("\n", " ").replace("\r", " ")`

AlbertvanHouten: The problem was that not all services initialized their loggers from here. I have rewritten this logic to always depend on geti_logger_tools for logger initialization. It should be fixed now but I will built an env and verify tomorrow.

interactive_ai/services/auto_train/app/controller.py
```diff
@@ -27,7 +27,7 @@
 from iai_core.utils.feature_flags import FeatureFlagProvider
 from iai_core.utils.time_utils import now
 
-logger = logging.getLogger(__name__)
+logger = initialize_logger(__name__)
```
leoll2: You don't have to call `initialize_logger` for every single file. `logging.basicConfig()` should be only called once per process, right after it starts. If you mount the sanitize filter on the root logger, I believe it will apply to all loggers down in the hierarchy.

leoll2 (changes requested):

interactive_ai/services/auto_train/app/controller.py
AlbertvanHouten: I have tried appending the filter to the root logger, but this doesn't propagate to any logger initialized afterward.

AlbertvanHouten: image

leoll2 (dismissed):

leoll2 (dismissed):

leoll2 (approved):

PR #517, opened 2025-06-23
feat: add component-based configuration for Geti CI

mpryahin:
## 📝 Description

- Add .github/components-config.yml for per-component build settings
- Create a reusable action for reading component settings
- Replace hardcoded component logic with configuration-driven approach

This allows developers to configure build behaviour per component instead of hardcoding settings in workflows, improving maintainability and build performance.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/.github/actions/read-component-config/action.yaml
+++ b/.github/actions/read-component-config/action.yaml
@@ -24,14 +24,17 @@ runs:
     - name: Read component configuration
       id: config
       shell: bash
+      env:
+        COMPONENT_DIR: ${{ inputs.component_dir }}
+        CONFIG_FILE: ${{ inputs.config_file }}
       run: |
         
         get_config_value() {
-          COMPONENT='${{ inputs.component_dir }}' KEY="$1" yq eval -r '
+          COMPONENT="$COMPONENT_DIR" KEY="$1" yq eval -r '
           .defaults as $defaults |
           (.components[env(COMPONENT)] // {}) as $component |
           ($defaults * $component)[env(KEY)]
-          ' ${{ inputs.config_file }}
+          ' "$CONFIG_FILE"
         }
 
         cleanup_type=$(get_config_value "cleanup_type")
@@ -41,7 +44,7 @@ runs:
         echo "integration_builder=$integration_builder" >> $GITHUB_OUTPUT
         
         echo "=== Component Configuration ==="
-        echo "Component: ${{ inputs.component_dir }}"
+        echo "Component: $COMPONENT_DIR"
         echo "cleanup_type: $cleanup_type"
         echo "integration_builder: $integration_builder"
         echo "==============================="
--- a/.github/actions/read-component-config/action.yaml
+++ b/.github/actions/read-component-config/action.yaml
@@ -14,7 +14,7 @@ outputs:
   cleanup_type:
     description: 'Type of cleanup to perform'
     value: ${{ steps.config.outputs.cleanup_type }}
-  integration_tests_builder_container:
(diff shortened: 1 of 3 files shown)


.github/components-config.yaml
```diff
@@ -0,0 +1,16 @@
+# Component Build Configuration
+#
+# This file optimizes CI/CD builds by configuring settings per component.
+# Components not listed here use the 'defaults' section below.
+components:
+  interactive_ai/workflows/otx_domain/trainer/otx_v2/xpu:
```
AlbertvanHouten, line 6: Could you also add otx_v2/gpu here?

.github/actions/read-component-config/action.yaml
```diff
@@ -0,0 +1,48 @@
+name: 'Read Component Configuration'
+description: 'Read component configuration from YAML file'
+author: 'Your Team'
```
AlbertvanHouten:
```suggestion
```

.github/components-config.yaml
mpryahin: done!

AlbertvanHouten (dismissed):

.github/workflows/component.yml
```diff
@@ -96,13 +104,15 @@ jobs:
 
       - name: Integration testing
         run: |
-          if [[ "${COMPONENT_DIR}" == "interactive_ai/workflows/geti_domain/project_ie" ]]; then
+          if [[ "${{ steps.component-config.outputs.integration_builder }}" == "false" ]]; then
```
MarkRedeman, line 107: In the [UI's Makefile](https://github.com/open-edge-platform/geti/blob/main/web_ui/Makefile#L28) we overwrite the builder step like,
```
builder:
	@echo "Not applicable"
```

This allows us to skip / ignore these steps from the component configuration, instead of the CI configuration.
Could we do something similar here (allowing us to remove this if statement)? So instead of doing an `@echo "Not applicable"` the target could run the provided arguments instead of forwarding them to a docker builder?

mpryahin: good idea, but in the current case we are just configuring runtime for CI - not skipping tests. If we remove these overwritten targets from web-ui makefile, then local builds will fail when they try to execute integration tests.

.github/workflows/component.yml
```diff
@@ -61,10 +61,18 @@ jobs:
           persist-credentials: false
           ref: ${{ inputs.ref || '' }}
 
+      - name: Read component config
```
mramotowski, line 64: Wouldn't it be easier to create 2 global vars/envs with list of components and then use [contains](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#contains) to check if it in the string/list? Less code and same readability.

AlbertvanHouten (approved):

.github/workflows/component.yml
mpryahin: Thanks, Maciej! That would only address conditional cleanup. Since we need more per-component options, it would force us to create even more variables, and enabling a setting across many components would clutter the workflow logic. By moving the configuration into a file, we follow [the open/closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle): our component workflow remains open for extension but closed for modification.

leoll2 (approved):

PR #544, opened 2025-06-25
QueryClientProvider refactor

jpggvilaca:
## 📝 Description

Why did this come up? I wanted to decouple notifications from a few places, which let me to update the mutation cache, which led to tests failing, which led to realising we were using different clients, which might be dangerous since then we'll be testing something different from what our main app uses. And here we are.
QueryClient now triggers notifications instead of each hook doing it. Which means the tests need the correct query client to test the existence of those notifications.

Note: I will wait for the release to merge this.

* `RequiredProvidersRender` will now use the actual query client we use for the geti app
* Extract query client creation to its own function
* Add mutationCache to our query client to inherit the same behavior we already have with queryCache (to trigger a notification onError)

For the next PRs:
* Leverage the query and mutation cache and get rid of all `addNotification` (non-custom ones) from all consumers.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/core/jobs/hooks/utils.test.tsx
+++ b/web_ui/src/core/jobs/hooks/utils.test.tsx
@@ -1,7 +1,7 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-import { QueryClient } from '@tanstack/react-query';
+import { useQueryClient } from '@tanstack/react-query';
 import { waitFor } from '@testing-library/react';
 
 import { getMockedWorkspaceIdentifier } from '../../../test-utils/mocked-items-factory/mocked-identifiers';
@@ -36,32 +36,39 @@ describe('Use jobs hook utils', () => {
     });
 
     it('Should not invalidate balance if feature flag is disabled', async () => {
-        const queryClient = new QueryClient();
-        queryClient.invalidateQueries = jest.fn();
+        const queryClient = renderHookWithProviders(useQueryClient);
+        queryClient.result.current.invalidateQueries = jest.fn();
+
         renderHookWithProviders(
-            () =>
-                useInvalidateBalanceOnNewJob(
+            () => {
+                return useInvalidateBalanceOnNewJob(
                     workspaceIdentifier,
                     getMockedResponse([getMockedJob({ cost: { leaseId: '123', requests: [], consumed: [] } })]),
                     { jobState: JobState.SCHEDULED }
-                ),
+                );
+            },
             {
-                providerProps: { queryClient, featureFlags: { FEATURE_FLAG_CREDIT_SYSTEM: false } },
+                providerProps: { featureFlags: { FEATURE_FLAG_CREDIT_SYSTEM: false } },
             }
         );
 
         await waitFor(() => {
-            expect(queryClient.invalidateQueries).not.toHaveBeenCalled();
+            expect(queryClient.result.current.invalidateQueries).not.toHaveBeenCalled();
         });
     });
 
     it('Should invalidate balance if feature flag is enabled', async () => {
-        const queryClient = new QueryClient();
-        queryClient.invalidateQueries = jest.fn();
+        let invalidateSpy: jest.SpyInstance | undefined;
+
         const { rerender } = renderHookWithProviders(
-            ({ jobs }) => useInvalidateBalanceOnNewJob(workspaceIdentifier, getMockedResponse(jobs), {}),
+            ({ jobs }) => {
+                const queryClient = useQueryClient();
+                invalidateSpy = jest.spyOn(queryClient, 'invalidateQueries');
+
+                return useInvalidateBalanceOnNewJob(workspaceIdentifier, getMockedResponse(jobs), {});
+            },
             {
-                providerProps: { queryClient, featureFlags: { FEATURE_FLAG_CREDIT_SYSTEM: true } },
+                providerProps: { featureFlags: { FEATURE_FLAG_CREDIT_SYSTEM: true } },
                 initialProps: {
                     jobs: [
                         getMockedJob({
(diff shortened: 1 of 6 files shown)


web_ui/src/core/jobs/hooks/utils.test.tsx
```diff
@@ -36,32 +36,39 @@ describe('Use jobs hook utils', () => {
...
 
     it('Should invalidate balance if feature flag is enabled', async () => {
-        const queryClient = new QueryClient();
-        queryClient.invalidateQueries = jest.fn();
+        let invalidateSpy: jest.SpyInstance | undefined;
+
         const { rerender } = renderHookWithProviders(
-            ({ jobs }) => useInvalidateBalanceOnNewJob(workspaceIdentifier, getMockedResponse(jobs), {}),
+            ({ jobs }) => {
+                const queryClient = useQueryClient();
```
jpggvilaca: we call `useQueryClient` inside the wrapper to get access to the client from the tree mounted by this test

web_ui/src/notification/notification.component.tsx
```diff
@@ -24,7 +24,7 @@ type addToastNotificationProps = Omit<NotificationToastProps, 'remove'> & {
     placement?: NOTIFICATION_CONTAINER;
 };
 
-interface addNotificationProps {
+export interface AddNotificationProps {
```
jpggvilaca, line 27: no big changes here, i just capitalised this, OCD

web_ui/src/test-utils/required-providers-render.tsx
```diff
@@ -26,70 +29,83 @@ interface RequiredProvidersProps extends Partial<ApplicationServicesContextProps
...
         <Suspense fallback={<IntelBrandedLoading />}>
             <Router initialEntries={initialEntries}>
-                <AuthProvider>
-                    <NotificationProvider>
-                        <Notifications />
-                        <QueryClientProvider client={queryClient ?? prefilledQueryClient}>
-                            <ThemeProvider theme={defaultTheme}>
+                <NotificationProvider>
+                    <Notifications />
+                    <PrefilledQueryClientProvider featureFlags={featureFlags} profile={profile}>
```
jpggvilaca, line 95: this is the main change

web_ui/src/providers/query-client-provider/query-client-provider.component.tsx
```diff
@@ -46,37 +106,12 @@ export const QueryClientProvider = ({
...
-                    if (query.meta.notifyOnError === true) {
-                        notify.current({
-                            message: typeof message === 'string' ? message : getErrorMessage(error),
-                            type: NOTIFICATION_TYPE.ERROR,
-                        });
-                    }
-                }
+        if (customQueryClient) {
+            return customQueryClient;
+        }
```
MarkRedeman: This feels a bit odd to me, if a user needs a different query client then they can render the provider manually,
```
        <TanstackQueryClientProvider client={customQueryClient}>
            // ...
        </TanstackQueryClientProvider>
```

instead of relying on this wrapper.

web_ui/src/test-utils/required-providers-render.tsx
```diff
@@ -26,70 +29,83 @@ interface RequiredProvidersProps extends Partial<ApplicationServicesContextProps
...
+const PrefilledQueryClientProvider = ({
+    children,
+    featureFlags,
+    profile,
+}: {
+    children: ReactNode;
+    featureFlags?: CustomFeatureFlags;
+    profile?: OnboardingProfile | null;
+}) => {
+    const { addNotification } = useNotification();
```
MarkRedeman, line 43: Do we need this notification behaviour in our tests? If we verify that `createGetiQueryClient` works as expected with a single unit test then I don't think we really need it here.

MarkRedeman (commented):
I'm not sure if I fully understand the intend of this PR. I'll try to come back to this later

web_ui/src/test-utils/required-providers-render.tsx
```diff
@@ -26,70 +29,83 @@ interface RequiredProvidersProps extends Partial<ApplicationServicesContextProps
...
-                <AuthProvider>
-                    <NotificationProvider>
-                        <Notifications />
-                        <QueryClientProvider client={queryClient ?? prefilledQueryClient}>
-                            <ThemeProvider theme={defaultTheme}>
+                <NotificationProvider>
+                    <Notifications />
+                    <PrefilledQueryClientProvider featureFlags={featureFlags} profile={profile}>
+                        <ThemeProvider theme={defaultTheme}>
+                            <Suspense fallback={<IntelBrandedLoading />}>
```
jpggvilaca, line 97: Why the extra suspense you might ask? Good question:

First, it's just for the tests so no danger there, secondly:

From google/copilot/chatgpt/testing:
"Sometimes, due to React’s rendering order or test environment quirks, the outer <Suspense> isn’t "ready" to catch suspensions from providers/components deep in the tree.
An inner <Suspense> ensures that any suspension in that subtree is always caught, preventing errors and showing the fallback as intended."

web_ui/src/providers/query-client-provider/query-client-provider.component.tsx
jpggvilaca: True, previous implementation quirks. Got rid of it now.

web_ui/src/test-utils/required-providers-render.tsx
jpggvilaca: we can also just pass the ref. But what you mean here is that we dont need notification stuff on the tests, but i dont think that is true. We do need it and we test a lot of notifications. Once we remove the "addNotification(...)" from the consumers (since our QC will trigger them by default) we will still be able to test the notification presence.

web_ui/src/core/jobs/hooks/utils.test.tsx
```diff
@@ -36,32 +36,39 @@ describe('Use jobs hook utils', () => {
...
         await waitFor(() => {
-            expect(queryClient.invalidateQueries).not.toHaveBeenCalled();
+            expect(queryClient.result.current.invalidateQueries).not.toHaveBeenCalled();
         });
     });
 
     it('Should invalidate balance if feature flag is enabled', async () => {
-        const queryClient = new QueryClient();
-        queryClient.invalidateQueries = jest.fn();
+        let invalidateSpy: jest.SpyInstance | undefined;
```
jpggvilaca: Discussed offline about these. I just "refactored" the tests to make it work with the new approach, but fundamentally they are not great tests in the sense that they test if react-query works instead of testing our app behavior AFTER react-query acts. So we should check if , once we do X, then Y happens, not that react-query called its internals correctly.

pplaskie (dismissed):

web_ui/src/providers/query-client-provider/query-client-provider.component.tsx
```diff
@@ -25,10 +26,67 @@ declare module '@tanstack/react-query' {
...
 }
 
+export const createGetiQueryClient = ({
+    defaultQueryOptions,
+    addNotification,
+}: {
+    defaultQueryOptions?: DefaultOptions;
+    addNotification: (props: AddNotificationProps) => string;
+}): QueryClient => {
+    const notify = { current: addNotification };
```
dwesolow, line 41: **Question:** why do we wrap `addNotification` with the object instead of calling in `addNotification` in line 49?

jpggvilaca: I did not made changes to this but I believe it's to keep track of the reference to the function? Kind of a custom `useCallback`. Because it the notification provider rerenders, the `addNotification` reference would differ from the one created when the query cache was created. But this is just my assumption i am not 100% sure

jpggvilaca: from line 99

```
// We want to make sure that query cache always uses the latest notification handler,
// but we don't want to reset the cache every time this happens so we keep a ref instead
```

dwesolow: Ah, I see. We should get `notify` that has stable reference instead of `addNotification` from the parent. Doing `const notify = { current: addNotification }` does not make the reference stable :).

jpggvilaca: Exactly, the "stable" part is being taken care of my the useRef + useLayoutEffect. Not stable but always the latest

dwesolow: There is no reason to create `notify` variable, we can just use `addNotification` from the parent.

web_ui/src/core/jobs/hooks/utils.test.tsx
```diff
@@ -29,39 +32,50 @@ const getMockedResponse = (jobs: Job[]) => ({
...
 const workspaceIdentifier = getMockedWorkspaceIdentifier({ workspaceId: 'workspaceId' });
+const mockSetInvalidateQueries = jest.fn();
+
+const queryClient = createGetiQueryClient({
+    addNotification: jest.fn(),
+});
+queryClient.invalidateQueries = mockSetInvalidateQueries;
+
+const wrapper = ({ children }: { children: ReactNode }) => (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
```
dwesolow, line 43: I think instead of creating second `QueryClientProvider` instance we could just leave the previous implementation of the `RequiredProvidersProps` which accepts optionally the `queryClient`. In case we want to mock `queryClient`, we create it using `createGetiQueryClient` and pass it to the hook/provider. For instance: `useInvalidateBalanceOnNewJob(..., { providerProps: { queryClient }})`.

PR #585, opened 2025-06-30
Remove logging config dir

AlbertvanHouten:
## 📝 Description
Remove logging config dir and replace it with LOG_LEVEL environment variable where applicable. The logging config dir was used to store logger config setting, which only held the log level. This made it very hard to change the log level since you would have to go into each pod, find this file, and then update the log level. Some of these log configs were also read only, making it not possible to change them. This PR adds a LOG_LEVEL environment variable to the deployment of applicable services, allowing for easier modification of the log level. 

https://jira.devtools.intel.com/browse/ITEP-70813

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


leoll2:
> This PR adds a LOG_LEVEL environment variable to the deployment of applicable services, allowing for easier modification of the log level.

How can you dynamically update the log level of an existing pod with the new system?

AlbertvanHouten:
> > This PR adds a LOG_LEVEL environment variable to the deployment of applicable services, allowing for easier modification of the log level.
> 
> How can you dynamically update the log level of an existing pod with the new system?

You can update the deployment of the pod and set the LOG_LEVEL env var to the desired value. This will cause the pods to restart and reflect the new log level.

leoll2 (approved):

itallix (approved):
LGTM!

platform/services/account/app/common/utils/logger.go
```diff
@@ -32,18 +29,12 @@ func syncLogLevel(loggerLevel *zap.AtomicLevel) {
 		"error":   zapcore.ErrorLevel,
 	}
 	for {
-		loggingConfigDir := os.Getenv(loggingConfigDirEnvVarName)
```
jcchr, line 35: The idea behind using this file to store log's level was to allow application to react to change of this level without a need of restarting pods - as only when you mount something as a file - its content is changed when the source is changed. Thanks this application may react if you change the source log level.
It doesn't work for env variables - as they remain unchanged for the whole life of a container - so to apply log's level change there - you need to restart the pod.
Thus I would suggest to stick to the current mechanism.

jcchr (changes requested):

platform/services/account/app/common/utils/logger.go
AlbertvanHouten: I've removed the old logic, as this feature was running a constant loop to check if the config file was changed, which is not ideal. With the new logic, if you update the environment variable in the deployment, the pods will also automatically restart. On top of that, some old config files were mounted in read only file systems, and they are not as easy to change as an environment variable, therefore I refactored the approach to log level changes. 

I also don't think that restart of pod is a problem. You will not get the different log level output from before when changing the log level.

jcchr: I still think the previous mechanism - if implemented properly - should be more convenient for users. After your change we need to modify manually all deployments where we need to change the log level. Previously - we had to modify this log level only in one place - and it affected all components which had this logic implemented. 
So maybe it would be better to stick to files - but refactor the mechanism to make it better?

AlbertvanHouten: I agree, I will revert the changes.

PR #619, opened 2025-07-03
ITEP-66305 - Move grabcut to smart-tools

### LINKED ISSUE #799: Move Grabcut

jpggvilaca:
## 📝 Description

* Move grabcut to smart-tools package
* Update interface, worker and consumer hooks

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/packages/smart-tools/src/grabcut/grabcut.ts
```diff
@@ -0,0 +1,327 @@
...
+
+class Grabcut {
+    img: OpenCVTypes.Mat;
+    mask: OpenCVTypes.Mat;
+    bgdModel: OpenCVTypes.Mat;
+    fgdModel: OpenCVTypes.Mat;
+
+    constructor(private CV: OpenCVTypes.cv) {}
+
+    loadImage(imageData: ImageData) {
```
jpggvilaca, line 19: [Note - only difference in this class is this function] I used the same strategy from watershed here, instead of having the imageData on the constructor, i added a new method responsible for initialiasing all this image data. Simplifying the constructor and make it consistent across other tools.

web_ui/src/webworkers/grabcut.worker.ts
```diff
@@ -1,321 +1,15 @@
...
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-//  Dependencies get bundled into the worker
+import { buildGrabcutInstance } from '@geti/smart-tools';
+import { expose, proxy } from 'comlink';
 
-import { approximateShape, OpenCVLoader } from '@geti/smart-tools';
-import { expose } from 'comlink';
-import type OpenCVTypes from 'OpenCVTypes';
+const WorkerApi = {
```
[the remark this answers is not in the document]
jpggvilaca: False positive

web_ui/packages/smart-tools/src/grabcut/grabcut.ts
MarkRedeman: One thing to check: if a user would do,
```
worker.loadImage(firstImage);
worker.loadImage(secondImage);
```

then in that case we don't do `this.mask.delete()` so that could introduce memory leaks within CV.

jpggvilaca: Valid concern, ill add a safeguard on the tool side

jpggvilaca: Updated

web_ui/packages/smart-tools/src/grabcut/grabcut.ts
```diff
@@ -307,10 +309,25 @@ class Grabcut {
     }
 
     cleanModels(): void {
-        this.img?.delete();
-        this.mask?.delete();
-        this.bgdModel?.delete();
-        this.fgdModel?.delete();
+        if (this.img) {
```
jpggvilaca, line 312: not only do we need to delete the mask but also to get rid of the reference (i was getting an error "Mask is already deleted!") so it seems that this more verbose approach is safer

web_ui/src/pages/annotator/hooks/use-grabcut.hook.ts
```diff
@@ -19,36 +23,52 @@ interface useGrabcutProps {
...
+                return Promise.reject(new Error('Could not run Grabcut'));
             }
         },
 
         onError: showNotificationError,
 
-        onSuccess: (data: Polygon, variables: GrabcutData) => {
-            onSuccess(data, variables);
+        onSuccess: (data: ToolPolygon, variables: GrabcutData) => {
+            const convertedPolygon = convertToolShapeToGetiShape(data) as Polygon;
```
dwesolow: I spent some time to check if we can make conversion without type assertion and I managed to make it work. Feel free to take a look, we gain type safety and we'd not need to assert if using other shapes.
```
export function convertToolShapeToGetiShape(shape: ToolPolygon): Polygon;
export function convertToolShapeToGetiShape(shape: ToolRect): Rect;
export function convertToolShapeToGetiShape(shape: ToolRotatedRect): RotatedRect;
export function convertToolShapeToGetiShape(shape: ToolCircle): Circle;

export function convertToolShapeToGetiShape(shape: SmartToolsShape): Shape {
...
}
```

jpggvilaca: Thanks for taking the time! Yeah i looked into it, overloading seems to benefit us here, it was already starting to be annoying and I know we're gonna keep using these util. tyty!

web_ui/src/pages/annotator/tools/ssim-tool/util.ts
```diff
@@ -105,7 +105,7 @@ export const convertRunSSIMPropsToToolRunSSIMProps = (runSSIMProps: RunSSIMProps
             shapeType: 'rect',
         },
         shapeType: convertGetiShapeTypeToToolShapeType(runSSIMProps.shapeType),
-        existingAnnotations: runSSIMProps.existingAnnotations.map(convertGetiShapeToToolShape),
+        existingAnnotations: runSSIMProps.existingAnnotations.map((shape) => convertGetiShapeToToolShape(shape)),
```
dwesolow: **NIT-PICK:** Previous solution should work as well.

web_ui/src/pages/annotator/tools/utils.ts
```diff
@@ -434,7 +441,12 @@ export const convertGetiShapeTypeToToolShapeType = (shapeType: ShapeType): Smart
     }
 };
 
-export const convertToolShapeToGetiShape = (shape: SmartToolsShape): Shape => {
+export function convertToolShapeToGetiShape(shape: ToolPolygon): Polygon;
+export function convertToolShapeToGetiShape(shape: ToolRect): Rect;
+export function convertToolShapeToGetiShape(shape: ToolRotatedRect): RotatedRect;
+export function convertToolShapeToGetiShape(shape: ToolCircle): Circle;
+export function convertToolShapeToGetiShape(shape: SmartToolsShape): Shape;
```
dwesolow, line 448: This overload is not necessary because the interface of the function is the same like in line 449.

web_ui/src/pages/annotator/tools/utils.ts
```diff
@@ -465,9 +477,14 @@ export const convertToolShapeToGetiShape = (shape: SmartToolsShape): Shape => {
...
     }
-};
+}
 
-export const convertGetiShapeToToolShape = (shape: Shape): SmartToolsShape => {
+export function convertGetiShapeToToolShape(shape: Polygon): ToolPolygon;
+export function convertGetiShapeToToolShape(shape: Rect): ToolRect;
+export function convertGetiShapeToToolShape(shape: RotatedRect): ToolRotatedRect;
+export function convertGetiShapeToToolShape(shape: Circle): ToolCircle;
+export function convertGetiShapeToToolShape(shape: Shape): SmartToolsShape;
```
dwesolow, line 486: Same here.

web_ui/src/pages/annotator/tools/ssim-tool/util.ts
jpggvilaca: It did not, i tried ^^

web_ui/src/pages/annotator/tools/utils.ts
jpggvilaca: It is. When exporting overloaded functions TS only cares about the definitions, not the implementation. So it never sees the "general" case . From https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads , the implementation signature is not visible from the outside, only the overloads are

dwesolow: Ah, I was 100% sure when I implemented it it was working with implementation signature. My bad ;)

dwesolow (dismissed):

web_ui/src/pages/annotator/tools/ssim-tool/util.ts
jpggvilaca: I tried BEFORE i added the last generic signature, so now i reverted and it works without extending the map :) . Updated

dwesolow (approved):

web_ui/packages/smart-tools/src/grabcut/grabcut.ts
camiloHimura: We didn’t encounter this error before. Is it possible that it’s happening because this method is being called before the properties are initialized? I’m not against this approach; I just want to understand why it’s needed now.

BTW, if you check whether `this.img` is not empty before calling `this.cleanModels()` on line 20, would all these other changes still be necessary?

jpggvilaca: This happens now because i added a "loadImage" method here on the grabcut class. Which cleans the models always before creating new masks

camiloHimura (approved):

MarkRedeman (approved):

PR #647, opened 2025-07-08
ITEP-66305 - Move RITM tool to smart-tools package

### LINKED ISSUE #801: Move RITM

jpggvilaca:
## 📝 Description

* Move RITM tool to smart-tools package. Only 2 more tools to go.
* Remove ndarray and ndarray-ops dependency from main geti-app

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


camiloHimura (dismissed):

web_ui/packages/smart-tools/src/ritm/interfaces.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright (C) 2022-2025 Intel Corporation
+// LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
+
+import * as ort from 'onnxruntime-web';
```
dwesolow, line 4: **Question:** Did you agree with Mark that we should import from `onnxruntime-common`, didn't you?
**Suggestion:** We can import the things we need instead of the whole package, e.g. `import { type Tensor } from 'onnxruntime-common';`

web_ui/packages/smart-tools/src/ritm/models/index.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2022-2025 Intel Corporation
+// LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
+
+export const RITMModels = {
```
dwesolow: **Question:** Do we expect to have models? I believe that no then I think we don't need barrel file, we can just call this file as `models.ts`.

web_ui/packages/smart-tools/src/ritm/interfaces.ts
jpggvilaca: 1) On the comments here yes, but offline we discussed and i decided to revert to whatever it was before.
2) Yes, i had that before but Mark did not agree with that approach, since it doesnt align with the docs. Personally i think the docs are outdated and we could do this, but it's really not a big deal i dont think

web_ui/packages/smart-tools/src/ritm/models/index.ts
jpggvilaca: I dont think so. Yeah i agree, i'll update it on the next PR with inteligent scissors: )

web_ui/src/pages/annotator/hooks/use-interactive-segmentation.hook.ts
```diff
@@ -46,9 +47,9 @@ export const useInteractiveSegmentation = ({
     useEffect(() => {
         const loadWorker = async () => {
             if (worker) {
-                ritmInstance.current = await new worker.RITM();
+                ritmInstance.current = worker;
 
-                await ritmInstance.current?.load();
+                await ritmInstance.current!.load();
```
dwesolow: **Blocking:** I'd not use this operator but either use safer option `?.load()` or `ritmInstance.current !== null && ritmInstance.current.load()`.

jpggvilaca: I pushed the "!" here because you can clearly see 1 line above that worker exists, and the ref exists as well. But just to be safe ill update this, along with the "models.ts" change

web_ui/packages/smart-tools/src/ritm/interfaces.ts
dwesolow: 1. OK
2. I think in the docs they recommend es modules import but not exactly `import * as` but I agree it's ambiguous..

dwesolow (dismissed):

camiloHimura (dismissed):

pplaskie (approved):

dwesolow (approved):

PR #706, opened 2025-07-11
Make geti shape types compatible with smart tool types

MarkRedeman:
## 📝 Description

This PR changes the values of our `ShapeType` to be string values that are equal to the relevant string types used by our `@geti/smart-tool` package. This makes it easier for us to switch between Geti's type and the smart tool package type (which we don't want to make aware of our enums as that adds a hard dependency to Geti).

Two important notes:
- To make the circle types compatible I had to revert the `x` to `cx` change. Perhaps later we can change our types to use `cx` and `cy` everywhere for both circle and rotated rects
- The `Pose` shape isn't fully compatible as we don't want to have the smart tool package to be aware of our label logic. See the last commit for some details

## ✨ Type of Change

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/packages/smart-tools/src/shared/interfaces.ts
+++ b/web_ui/packages/smart-tools/src/shared/interfaces.ts
@@ -31,13 +31,9 @@ export interface Polygon {
     readonly shapeType: 'polygon';
 }
 
-interface KeypointNode {
-    readonly x: number;
-    readonly y: number;
-    readonly isVisible: boolean;
-}
 export interface Pose {
-    readonly points: KeypointNode[];
+    // NOTE: this is not fully compatible with Geti's type
+    readonly points: Point[];
     readonly shapeType: 'pose';
 }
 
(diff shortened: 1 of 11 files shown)


jpggvilaca (approved):

web_ui/packages/smart-tools/src/shared/interfaces.ts
```diff
@@ -31,7 +31,13 @@ export interface Polygon {
     readonly shapeType: 'polygon';
 }
 
-export type Shape = Rect | RotatedRect | Circle | Polygon;
+export interface Pose {
+    // NOTE: this is not fully compatible with Geti's type
```
dwesolow, line 35: **Question:** Did u add `Pose` to make typescript happy while mapping between geti and package shapes?
IMO we should not add `Pose` to smart tools since keypoint is not part of this package.
Thoughts?

MarkRedeman: Yes because otherwise the `Pose` shape from Geti won't be assignable to she `Shape` interface from the smart tool package.
I do agree it feels a bit icky, but this allows us to move the shape utility functions like `isPointInShape` to this package as well - perhaps later we can think of a better name than `smart-tools` as it's getting a few more responsibilities.

Some other thoughts I have is that we could move the logic from the selection and edit tool here as well.

dwesolow (approved):

PR #676, opened 2025-07-10
ITEP-66305 - Cleanup smart tools

jpggvilaca:
## 📝 Description

* Cleanup interfaces and migration code (Remove worker interface in favor of a generic BaseWorker and make sure all tools use Remote type)

## ✨ Type of Change

Select the type of change your PR introduces:

- [ ] 🐞 **Bug fix** – Non-breaking change which fixes an issue
- [ ] 🚀 **New feature** – Non-breaking change which adds functionality
- [x] 🔨 **Refactor** – Non-breaking change which refactors the code base
- [ ] 💥 **Breaking change** – Changes that break existing functionality
- [ ] 📚 **Documentation update**
- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Describe how the changes were tested and how reviewers can test them too:

- [x] ✅ Tested manually
- [ ] 🤖 Run automated end-to-end tests

## ✅ Checklist

Before submitting the PR, ensure the following:

- [x] 🔍 PR title is clear and descriptive
- [x] 📝 For internal contributors: If applicable, include the JIRA ticket number (e.g., ITEP-123456) in the PR **title**. Do **not** include full URLs
- [ ] 💬 I have commented my code, especially in hard-to-understand areas
- [ ] 📄 I have made corresponding changes to the documentation
- [ ] ✅ I have added tests that prove my fix is effective or my feature works

PR #674, opened 2025-07-10
ITEP-66305 - Move SegmentAnything tool to smart-tools package

jpggvilaca:
## 📝 Description

* Last tool (finally hum?)
* Next PR: Cleanup and improve typing

## ✨ Type of Change

Select the type of change your PR introduces:

- [ ] 🐞 **Bug fix** – Non-breaking change which fixes an issue
- [ ] 🚀 **New feature** – Non-breaking change which adds functionality
- [x] 🔨 **Refactor** – Non-breaking change which refactors the code base
- [ ] 💥 **Breaking change** – Changes that break existing functionality
- [ ] 📚 **Documentation update**
- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Describe how the changes were tested and how reviewers can test them too:

- [x] ✅ Tested manually
- [ ] 🤖 Run automated end-to-end tests

## ✅ Checklist

Before submitting the PR, ensure the following:

- [x] 🔍 PR title is clear and descriptive
- [x] 📝 For internal contributors: If applicable, include the JIRA ticket number (e.g., ITEP-123456) in the PR **title**. Do **not** include full URLs
- [ ] 💬 I have commented my code, especially in hard-to-understand areas
- [ ] 📄 I have made corresponding changes to the documentation
- [ ] ✅ I have added tests that prove my fix is effective or my feature works

PR #753, opened 2025-07-15
ITEP-66305 - Move all tool related utilities

jpggvilaca:
## 📝 Description

* Move vec2, math and rotated-rect utils
* Update all imports
* Fix Opencv type imports (not the final fix. I want to convert all those .d.ts files with "declare module..." into actual modules, so we can use import/export like it's 2025)
* Extend the exports from the smart-tools package, we can now do `import x from @geti/smart-tools/subpackage`, where subpackage can now be `opencv | types | utils | ritm`. Will add more later on as needed. I know there's mixed opinions about exposing the subpackages like this but this will allow for more flexibility when testing. Since we will be able to mock everything specifically.

After we merge this, I'll move the last "smart" tool with https://github.com/open-edge-platform/geti/pull/674 . And then further cleanup with https://github.com/open-edge-platform/geti/pull/676

Note: Reviewing commit by commit is much easier

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/packages/config/typescript/tsconfig.base.json
```diff
@@ -14,7 +14,7 @@
     "forceConsistentCasingInFileNames": true,
     "noFallthroughCasesInSwitch": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
jpggvilaca, line 17: this was necessary to actually make the "exports" on `package.json` work.

If you want know more check https://www.typescriptlang.org/tsconfig/#moduleResolution , but the important part is:

`'bundler' for use with bundlers. Like node16 and nodenext, this mode supports package.json "imports" and "exports", but unlike the Node.js resolution modes, bundler never requires file extensions on relative paths in imports.`

web_ui/packages/smart-tools/src/utils/wasm-utils.ts
```diff
@@ -4,27 +4,19 @@
...
-    ).toString(),
-    'ort-wasm-threaded.wasm': new URL(
-        '../../../../node_modules/onnxruntime-web/dist/ort-wasm-threaded.wasm',
-        import.meta.url
-    ).toString(),
-    'ort-wasm-simd-threaded.wasm': new URL(
-        '../../../../node_modules/onnxruntime-web/dist/ort-wasm-simd-threaded.wasm',
-        import.meta.url
-    ).toString(),
+const getWasmPaths = async () => {
```
jpggvilaca: this change was necessary because if we don't lazy load the paths, then the URL executes immediately when the file is loaded, which causes a lot of issues when we're on the jest context. To circumvent that, it was recommended to move the calls to URL to a different file, and lazy load it to avoid jest executing this. Actual error: "import.meta cannot be executed outside of a module" .

Before this i also tried using `moduleNameWrapper` (because it executes before tests are initialized) but then i would need to mock ALL utils one by one, which broke hundreds of tests.

MarkRedeman: I don't fully understand why this is necessary now (and not before?) or why we only need to do this when getting the wasm paths and not when getting the paths for the ritm or sam models.
Can you elaborate on that?

web_ui/packages/smart-tools/package.json
```diff
@@ -25,7 +26,10 @@
   },
   "exports": {
     ".": "./index.ts",
-    "./opencv": "./src/opencv/interfaces/index.d.ts"
+    "./opencv": "./src/opencv/interfaces/index.d.ts",
+    "./utils": "./src/utils/index.ts",
+    "./types": "./src/shared/index.ts",
+    "./ritm": "./src/ritm/index.ts"
```
MarkRedeman, line 32: why do we need to do this only for ritm and not the other tools?

jpggvilaca: We dont need do to it for any tool. It was just to serve as an example. Then later i'll add the rest as we discussed 500prs ago (when i first started moving these). Also i didnt want to make all those changes on this PR

web_ui/packages/smart-tools/src/utils/wasm-utils.ts
jpggvilaca: Because now the utils are on the package, along with unit tests, which brings jest into the mix. Not only geti now imports these utils, but we also test these utils on the smart-tools package. And when jest starts to load smart-tools, it also loads this wasm-utils.ts file, which brings all sorts of issues. Before we didnt have any tests on the package, nor did we import "session" or wasm stuff from our unit tests

web_ui/packages/smart-tools/src/utils/math.ts
```diff
@@ -185,15 +183,15 @@ export const pointInPolygon = (polygon: Polygon, point: Point): boolean => {
 
 export const isPointInShape = (shape: Shape, point: Point): boolean => {
     switch (shape.shapeType) {
-        case ShapeType.Polygon:
+        case 'polygon':
```
dwesolow, line 186: Why these changes to `shapeType`? Is it because `ShapeType` enum is not in the smart tools package?

jpggvilaca: Because https://github.com/open-edge-platform/geti/pull/706 ^^ . Maybe i can use the "ShapeType" from geti tools package directly but i didnt waste too much time on that

web_ui/packages/smart-tools/src/ritm/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (C) 2022-2025 Intel Corporation
+// LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
+
+export { buildRITMInstance, RITM, RITM_TEMPLATE_SIZE } from './ritm';
+export { RITMModels } from './models/models';
```
MarkRedeman: I don't think we should expose this since it's not something we would want the consumer of the package to use directly: only `buildRITMInstance` and the interface types belong to our public interface.

web_ui/packages/smart-tools/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "@geti/config/typescript",
+  "compilerOptions": {
+    "types": ["node", "jest"],
```
dwesolow: why this was necessary? AFAIR types that starts with `@types/` are automatically added under the hood.

jpggvilaca: it was in order to use `process.env` on the "hacky" solution. But since it got removed we can get rid of these 2

jpggvilaca: Removed

web_ui/packages/smart-tools/src/utils/wasm-utils.ts
dwesolow: After the latest change I believe this file can be removed, right?

jpggvilaca: Correct, removed. Thanks!

dwesolow (approved):

camiloHimura (approved):

MarkRedeman (approved):

PR #760, opened 2025-07-15
:chore: - Refactor FullscreenAction component

jpggvilaca:
## 📝 Description

The goal of the PR is to make this fullscreen UI more reusable

* Decouple chart dependencies from component
* Replaced all occurrences
* Add 1 more unit test

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/src/pages/project-details/components/project-model/model-statistics/utils.tsx
```diff
@@ -23,15 +23,15 @@ export const getModelStatisticPresentation = (
         case 'text':
             return <TrainingModelTextChart value={statistic.value} header={statistic.header} {...options} />;
         case 'line':
-            return <TrainingModelLineChart {...statistic} {...options} />;
+            return <TrainingModelLineChart {...statistic} key={statistic.key} {...options} />;
```
dwesolow: Why this change?

web_ui/packages/ui/src/fullscreen-action/fullscreen-action.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+                    <div data-testid={'test-fullscreen'} />
+                </FullscreenAction>
+            </ThemeProvider>
+        );
+
+        expect(screen.queryByTestId('test-fullscreen')).not.toBeInTheDocument();
+        const expandButton = screen.getByRole('button', { name: /Open in fullscreen/ });
+        fireEvent.click(expandButton);
+
+        const dialog = screen.queryByRole('dialog');
```
dwesolow: np. `getByRole`

web_ui/src/pages/project-details/components/project-model/model-statistics/utils.tsx
jpggvilaca: Because react yells at us if we try to spread a prop with `key`.

web_ui/packages/ui/src/fullscreen-action/fullscreen-action.test.tsx
jpggvilaca: updated

dwesolow (dismissed):

web_ui/src/pages/project-details/components/project-model/model-statistics/utils.tsx
MarkRedeman: Can you check this again? I don't see why we would need a `key` prop here.

web_ui/src/pages/project-details/components/project-model/model-statistics/training-model-matrix/training-model-matrix.component.tsx
```diff
@@ -64,13 +64,20 @@ export const TrainingModelMatrix = ({
...
+                                graphBackgroundColor={'gray-100'}
+                            />
+                        }
                         title={header}
-                        downloadableData={{ type: 'matrix', data: confusionMatrixData }}
                     >
                         <ConfusionMatrixZoom
                             title={header}
                             {...confusionMatrixData}
+                            key={confusionMatrix.key}
```
MarkRedeman: Why do we need to add this key here? It's not part of a map so we shouldn't need to add a key iirc.
(Same for the line below)

dwesolow: I think it has the issue because `key` is a `react`'s prop and `key` is also the prop of the `ConfusionMatrixZoom`. The same name but different purpose. I remember I saw the same issue but in the a different place.

jpggvilaca: Correct, same for the other case above. I dont think it breaks anything, but the devtools was all red because of this. I can also revert this as it's out of the scope of this PR

camiloHimura (dismissed):

web_ui/src/pages/project-details/components/project-model/model-statistics/training-model-matrix/training-model-matrix.component.tsx
MarkRedeman: But the `ConfusionMatrixZoom` component isn't using any `key` prop, and it's not being used in a scenario where react requires it.
```typescript
export interface ConfusionMatrixProps {
    columnHeader: string;
    rowHeader: string;
    columnNames: string[];
    rowNames: string[];
    matrixValues: number[][];
    size: number;
    ratio?: number;
}

interface ConfusionMatrixZoomProps extends ConfusionMatrixProps {
    size: number;
    title: string;
}
```

Please check again and also do this for the other parts where `key` was added.
By adding key you can unintentionally trigger remounting of the components so we should avoid this when it's not needed.

dwesolow: image

Let's try not to spread parameters but only pass parameters that are needed (could we do it in a separate PR not to increase the size of this PR and revert this change?)

jpggvilaca: Agree. I will revert the change

jpggvilaca: Reverted

web_ui/src/pages/project-details/components/project-model/model-statistics/utils.tsx
jpggvilaca: Commented below, reverted.

camiloHimura (dismissed):

MarkRedeman (dismissed):

camiloHimura (approved):

dwesolow (approved):

PR #667, opened 2025-07-09
ITEP-71155 - Update training job flow

maxxgx:
## 📝 Description

JIRA: ITEP-71155

- *Geti -> OTX*: replace "config.json" with "config.yaml" containing yaml configuration, for example:
```yaml
job_type: train
model_manifest_id: Custom_Image_Classification_EfficinetNet-B0
hyperparameters:
  data_preparation: 
    augmentation:
      horizontal_flip: true,
      vertical_flip: false,
    ...
  training:
    max_epochs: 100,
    batch_size: 32,
    learning_rate: 0.0049
    ...
  evaluation:
    ...
export_models:
  - format: openvino
    output_model_id: 67a32c53d8c2d927812b0b3b
    precision: FP32
    with_xai: true
  - format: openvino
    output_model_id: 67a32c53d8c2d927812b0b3c
    precision: FP32
    with_xai: false
  - format: openvino
    output_model_id: 67a32c53d8c2d927812b0b3d
    precision: FP16
    with_xai: false
  - format: onnx
    output_model_id: 67a32c53d8c2d927812b0b3e
    precision: FP32
    with_xai: false
```
- *OTX -> Geti*: return `advanced_config.json` containing extra configuration used by OTX, which was NOT included in the input config.yaml. The “advanced” configuration parameters cannot be set by the user because they lack backend definitions and are therefore only used for visualization. Example:
```json
[
    {
        "key": "optimum_confidence_threshold",	 
        "name": "Optimum confidence threshold",
        "description": "The confidence threshold for ideal predictions",
        "value": 0.65
    }
]
```

Also resolves issues:
- https://github.com/open-edge-platform/geti/issues/877
- https://github.com/open-edge-platform/geti/issues/878

## ✨ Type of Change

Select the type of change your PR introduces:

- [ ] 🐞 **Bug fix** – Non-breaking change which fixes an issue
- [x] 🚀 **New feature** – Non-breaking change which adds functionality
- [ ] 🔨 **Refactor** – Non-breaking change which refactors the code base
- [ ] 💥 **Breaking change** – Changes that break existing functionality
- [ ] 📚 **Documentation update**
- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Describe how the changes were tested and how reviewers can test them too:

- [x] ✅ Tested manually
- [x] 🤖 Run automated end-to-end tests

## ✅ Checklist

Before submitting the PR, ensure the following:

[...]

PR #855, opened 2025-07-18
Modify OTX scripts and support OTX v2.5.0

### LINKED ISSUE #815: Modify Geti OTX scripts to be able to run training with new model manifests and OTX 2.5
OTX v2.5.0 introduces API changes, including a new model manifest format.
These changes break compatibility with OTXConfig and ConfigConverter. To ensure that training and optimization tasks continue to run properly with the updated training backend and configuration format, we need to apply the necessary adjustments accordingly.

kprokofi:
## 📝 Description

Provide a clear summary of the changes and the context behind them. Describe **what** was changed, **why** it was needed, and **how** the changes address the issue or add value.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist


interactive_ai/workflows/train/trainer/pyproject.toml
```diff
@@ -9,7 +9,7 @@ dependencies = [
     "numpy==1.26.4",
     "requests==2.32.3",
     "tqdm==4.67.1",
-    "otx[base]==2.4.6",
+    "otx@git+https://github.com/open-edge-platform/training_extensions@4107731720f53fc61c51d287a79a23d221c52e84",
```
leoll2: Can you keep the `[base]` modifier? Or is it no longer applicable since OTX 2.5?

interactive_ai/workflows/train/trainer/pyproject.toml
```diff
@@ -44,6 +44,8 @@ build-backend = "hatchling.build"
 [tool.hatch.build.targets.wheel]
 packages = ["trainer"]
 
+[tool.hatch.metadata]
+allow-direct-references = true
```
leoll2: I guess you added this because of the `@` version specifier in otx package.

There is an alternative way to reference a package at a specific revision:
```
dependencies = [
    "otx",
]

[tool.uv.sources]
otx = { git = "https://github.com/open-edge-platform/training_extensions.git", rev = "4107731720f53fc61c51d287a79a23d221c52e84" }
```
Instead of `rev` you can also use `branch = "release/2.5"`. I believe the solution based on `[tool.uv.sources]` is slightly preferable because more flexible and doesn't require direct references. ~Possibly, it doesn't even require you install git in the Dockerfile, but I didn't verify it.~ (EDIT: it still does)

interactive_ai/workflows/train/trainer/pyproject.toml
kprokofi: Yes, it is no longer needed. "pip install otx" is sufficient

interactive_ai/workflows/train/trainer/scripts/utils.py
```diff
@@ -154,93 +151,40 @@ def str2bool(value: str | bool) -> bool:
...
+    model_manifest_id: str
     hyper_parameters: dict
     export_parameters: list[ExportParameter]
     optimization_type: OptimizationType | None
-    sub_task_type: OTXTaskType | None
+    sub_task_type: str | None = None
 
     @classmethod
-    def from_json_file(cls, config_file_path: Path) -> OTXConfig:
+    def from_yaml_file(cls, config_file_path: Path) -> OTXConfig:
```
leoll2, line 161: Does OTX 2.5 still support the legacy JSON or was it completely removed in favor of the YAML?
Asking because in the backend we have a feature flag (FF_NEW_CONFIGURABLE_PARAMETERS) which controls a bunch of things, including reading the manifests and building the training config in YAML format. If 2.5 doesn't support JSON, then it implies that FF_NEW_CONFIGURABLE_PARAMETERS must be enabled as soon as we switch to 2.5.

interactive_ai/workflows/train/trainer/pyproject.toml
kprokofi: [removed [tool.hatch.metadata], updated installation using branch](https://github.com/open-edge-platform/geti/pull/855/commits/ea2af75cd05b4bb29356bf43f6c2b334a4a6b74c)

interactive_ai/workflows/train/trainer/scripts/utils.py
kprokofi: We don't support both Json and Yaml. So, starting from 2.5 we need to switch to Yaml format only. Is it OK?

kprokofi: Supporting both formats would lead to messy code with the current ConfigConverter implementation. I’d prefer to stick to a single format and remove the model templates from OTX.

leoll2: Yes it's okay from my point of view, as long as the existing projects and the other features keep working after the switch. Let's test it extensively next week.

leoll2:
I think we can close this PR in favor of #667

kprokofi:
> I think we can close this PR in favor of #667

I am using this branch right now to fix unit tests. I will close this PR after that and Max will merge my branch

PR #857, opened 2025-07-18
fix: Project creation template

dwesolow:
## 📝 Description

This PR:
- fixes jumping width effect when switching between different task types;
- adds animation when going to project name step;
- sets min height for a dialog so we don't have jumping height effect when switching between task types.

https://github.com/user-attachments/assets/59c5206b-0e40-4fcf-8c86-5a6fed2d5bb5

## ✨ Type of Change

- [X] 🐞 **Bug fix** – Non-breaking change which fixes an issue

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/create-project/components/card.component.tsx
+++ b/web_ui/src/pages/create-project/components/card.component.tsx
@@ -53,10 +53,10 @@ export const Card = ({
             {...pressProps}
             id={id}
             data-testid={id}
-            className={clsx({
-                [classes.card]: true,
+            className={clsx(classes.card, {
                 [classes.selected]: isSelected,
                 [classes.disabled]: isDisabled,
+                [classes.notSelected]: !isSelected,
             })}
             aria-label={title}
         >
--- a/web_ui/src/pages/create-project/new-project-dialog.component.tsx
+++ b/web_ui/src/pages/create-project/new-project-dialog.component.tsx
@@ -101,7 +101,7 @@ const NewProjectDialogInner: FC<NewProjectDialogInnerProps> = ({ onCloseDialog }
 
     return (
         <Dialog
-            height='100%'
+            minHeight={'52vh'}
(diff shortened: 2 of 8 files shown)


web_ui/src/pages/create-project/components/card.component.tsx
```diff
@@ -53,10 +53,10 @@ export const Card = ({
             {...pressProps}
             id={id}
             data-testid={id}
-            className={clsx({
-                [classes.card]: true,
+            className={clsx(classes.card, {
                 [classes.selected]: isSelected,
                 [classes.disabled]: isDisabled,
+                [classes.notSelected]: !isSelected,
```
jpggvilaca, line 59: question: isnt this the default? Can we add the notSelected styles to card styles?

web_ui/src/pages/create-project/new-project-dialog.component.tsx
```diff
@@ -101,7 +101,7 @@ const NewProjectDialogInner: FC<NewProjectDialogInnerProps> = ({ onCloseDialog }
 
     return (
         <Dialog
-            height='100%'
+            minHeight={'52vh'}
```
jpggvilaca, line 104: Is this value consistent with other dialogs?

dwesolow: No, I adjusted this value only for this dialog because I didn't want it to have more space than necessary.

web_ui/src/pages/create-project/components/card.component.tsx
dwesolow: No, it was not default. When I tried to apply it to `.card` then I'd need to do `!important` for `.selected` so decided to do it as a separate class.

camiloHimura (approved):

pplaskie (approved):

PR #946, opened 2025-07-31
BDD tests for project export

JortBergfeld:
## 📝 Description

Add BDD tests for project export and the include_models parameter.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [x] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist

- [x] ✍️ PR description clearly explains the changes and their reason

### DIFF
--- a/interactive_ai/services/api/schemas/project_import_export/responses/export_project_metadata.yaml
+++ b/interactive_ai/services/api/schemas/project_import_export/responses/export_project_metadata.yaml
@@ -3,8 +3,18 @@ type: object
 description: Information about the project export operation
 properties:
   project:
-    description: Information about the project being export
+    description: Information about the project being exported
     $ref: 'project.yaml'
+  include_models:
+    description: Enum indicating which models are included in the exported zip
+    type: string
+    enum:
+      - all
+      - none
+      - latest_active
   download_url:
     type: string
     description: URL to download the exported project archive
+  size:
+    type: integer
+    description: Size of the exported project zip file
--- a/interactive_ai/tests/e2e/features/steps/project_import_export.py
+++ b/interactive_ai/tests/e2e/features/steps/project_import_export.py
@@ -154,9 +154,11 @@ def step_when_user_requests_project_export(context: Context, included_models: st
     context.job_id = project_export_response.job_id
 
 
-@then("the exported project can be downloaded")
-def step_then_user_can_download_exported_project(context: Context) -> None:
+@then("the exported project can be downloaded and is {exported_project_size} MB")
+def step_then_user_can_download_exported_project(context: Context, exported_project_size: int) -> None:
     metadata = context.job_info.actual_instance.metadata
+    expected_size = int(exported_project_size) * 1000000
+    assert expected_size <= int(metadata.size) < expected_size + 1000000
     download_url = metadata.download_url
     export_id = download_url.split("/")[-2]
     project_import_export_api = context.project_import_export_api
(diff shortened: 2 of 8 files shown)


interactive_ai/tests/e2e/features/steps/project_import_export.py
```diff
@@ -140,3 +140,29 @@ def step_when_user_uploads_project_to_new_project(context: Context) -> None:
...
+    metadata = context.job_info.actual_instance.metadata
+    download_url = metadata.download_url
+    export_id = download_url.split("/")[-2]
+    project_import_export_api = context.project_import_export_api
+    project_import_export_api.download_exported_project(
+        organization_id=context.organization_id,
+        workspace_id=context.workspace_id,
+        project_id=context.project_id,
+        export_operation_id=export_id,
+    )
```
leoll2, line 170: Does this method actually download the file, or does it only return a presigned url?

leoll2: I wonder if we should make additional assertions, e.g.
- the downloaded project is a zip file
- the size is within some thresholds (e.g. with 'all' or 'latest_active' models it should be > than a few megabytes, while with 'none' it should be very light)

JortBergfeld: I added an assert to check the different sizes of the different options. Checking if it is an actual zip file takes more effort and may not be so important. I think in a later stage we should reimport the zip file to ensure the reimported project works as expected. A successful export job does not imply the zip file can be imported without error.

Note: I fixed the API schemas in another PR, but I needed that in this PR to check the size of the export. So I merged both branches now.

leoll2: The check below can become very flaky, as the exact archive size may oscillate as we modify the database schema to accommodate new features, or even because of changes in the OpenVINO model optimizer.

```gherkin
And the exported project can be downloaded and is <exported_project_size> MB
```

My proposal instead was to use ranges. Based on your numbers, a reasonable threshold could be 20MB, knowing that all model binaries always take more space than that.

interactive_ai/services/api/schemas/project_import_export/responses/export_project_metadata.yaml
```diff
@@ -3,8 +3,18 @@ type: object
...
+      - none
+      - latest_active
   download_url:
     type: string
-    description: URL to download the exported project archive
\ No newline at end of file
+    description: URL to download the exported project archive
+  size:
+    type: integer
+    description: Size of the exported project zip file
```
leoll2, line 20: What is this field?

JortBergfeld: This is the size of the download. It already existed before, but I just added it to the docs, since I was fixing the metadata docs for adding the `include_models` property anyway.

maxxgx (approved):

A-Artemis (approved):

interactive_ai/tests/e2e/features/steps/project_import_export.py
JortBergfeld: I don't think "(very) flaky" is the correct term here. We might have to adjust the test to changes, but it tests at the MB level, so we have 1 million bytes for minor DB changes. For OpenVINO optimizer changes, which should only occur at most 4 times a year, we might have to adjust the test, but that is normal. Even then, it would only occur if we change the BDD test data, because right now it simply exports the imported binaries from the BDD test data zip.

PR #1020, opened 2025-08-06
#1117 Update spectrum

### LINKED ISSUE #1117: Update react spectrum to the latest version

dwesolow:
## 📝 Description

This PR updates spectrum to the latest version.

Environment to test: https://10.55.252.150/.

https://react-spectrum.adobe.com/releases/2025-07-22.html

This PR includes several changes:
1. Fixes locators in the playwright tests - to make them really specific. In the past, when we had a button "Add" (1) which opened a modal that also had a button "Add" (2) -> 1 was not visible when dialog was displayed. After the upgrade both buttons are visible so we needed stricter locators.
2. Menus that used `isHover` using js to show/hide menu stopped working because of `isHovered` was toggled to off when menu appeared. I changed it so we use pure css for this.
3. When there is a popover inside popover, spectrum decided to add `inert` property that disables all the interactivity. It broke the context menu in the key point detection project template to delete the edge. I changed the implementation a bit to leverage the `createPortal`.

**Please test the changes locally.**

Closes #1117 

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [X] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

- [X] 🔍 PR title is clear and meaningful
- [X] ✍️ PR description clearly explains the changes and their reason
- [X] ✅ I have added tests that prove my fix is effective or my feature works


pplaskie (dismissed):

web_ui/src/pages/create-project/components/pose-template/canvas/cursor-context-menu/cursor-context-menu.component.tsx
```diff
@@ -63,3 +63,99 @@ export const CursorContextMenu = ({
         </OverlayContainer>
     );
 };
+
+/*return (
```
camiloHimura: leftover?

dwesolow: No, commented my attempts while trying to fix the issue after the update but they didn't work.
The issue is that spectrum decided to use `inert` instead of `aria-hidden` in ariaHideOutside. It means that when we have a overlay inside overlay, the nested overlay is not interactive. The solution for this would be to refactor code to use popover (?) but I didn't manage to make it work.

romanowska (dismissed):

web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
```diff
@@ -121,23 +78,8 @@ const Labels = ({ children, ...otherProps }: FlexProps) => {
...
-
-    return (
-        <>
-            <Flex {...otherProps} gridArea={'list-menu'} ref={containerRef}>
-                {children}
-            </Flex>
-            <InteractOutside containerRef={unwrapContainerRef} onClickOutside={() => setIsHovered(false)} />
-        </>
-    );
+const ListMenu = ({ children }: FlexProps) => {
```
MarkRedeman, line 81: Could you clarify the changes here? Perhaps we can inline this `Flex` component as it no longer has it's own behaviour? 

Edit: on second look: it seems like we now use css's over styles?
I'm wondering though why we had it like this in the first place @camiloHimura can you double check this?

web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
```diff
@@ -56,59 +46,49 @@ export const GridMediaItem = ({
...
-                        paddingY={'size-150'}
-                        paddingX={'size-75'}
-                        height={'size-500'}
-                        zIndex={10}
-                    >
-                        <MediaItemActions
-                            mediaItem={mediaItem}
-                            onSelectedMediaItemActionChange={setSelectedMediaItemAction}
-                            selectedMediaItemAction={selectedMediaItemAction}
+            <Flex
```
MarkRedeman, line 50: One thing to note here: I think here we were conditionally rendering based on `isHovered` and `isSelected` to improve the performance of the media grid.
One thing I've seen lately is that `Checkbox` can be quite slow to render which is more clear when you go to the media page and use many small thumbnails.

web_ui/src/pages/annotator/annotation/list-item-grid.component.tsx
dwesolow: Yes, we leverage css now.
There was an issue that when menu was about to open, element lost the pointer (onPointerOut) and `isHovered` was toggled off and on, which caused that menu was not visible (the same happened with the media item menu (three dots when hovering media item). I decided to use pure css (as we agreed, use css where we can, if not possible with css, use js).

web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
dwesolow: We use `isSelected` to change visibility of the element using css. Feel free to test the PR and please share if u find any regression :).

web_ui/src/pages/create-project/components/pose-template/canvas/cursor-context-menu/cursor-context-menu.component.tsx
dwesolow: Updated

camiloHimura:
This is not a good solution from a UX perspective. The delete icon appears quite far from the click point, making it unclear which element it belongs to. This is why I didn’t use `popover`, we don’t want the element to appear at the top, left, right, or bottom of the target, but rather close to the click event. We need to explore other options

image

dwesolow:
> This is not a good solution from a UX perspective. The delete icon appears quite far from the click point, making it unclear which element it belongs to. This is why I didn’t use `popover`, we don’t want the element to appear at the top, left, right, or bottom of the target, but rather close to the click event. We need to explore other options
> 
> image

I didn't find better approach with spectrum components. Please explore if you have some time.

web_ui/src/pages/project-details/components/project-media/grid-media-item.component.tsx
MarkRedeman: I've done some testing and it looks like the lag I was thinking of is not noticeable on your environment. So I'm fine with sticking to this simpler solution.

camiloHimura:
> I didn't find better approach with spectrum components. Please explore if you have some time.

I explored all possible options when I began working on that implementation and arrived at the current solution. Perhaps we should consider options outside spectrum

romanowska (dismissed):

web_ui/src/pages/create-project/components/pose-template/canvas/cursor-context-menu/cursor-context-menu.component.tsx
```diff
@@ -36,30 +30,45 @@ export const CursorContextMenu = ({
...
+                left={cursorPosition.x}
+                zIndex={100001}
+                backgroundColor={'gray-200'}
+                data-testid='position container'
+                {...overlayProps}
+            >
+                {children}
+            </View>
+        </div>,
+        document.querySelector('[data-testid="modal"]') as HTMLElement
```
camiloHimura: It took me a moment to realize this selector refers to the parent modal, not the newly created portal elements. Let’s move this selector into a variable that can be used both at line 33 and here. `PARENT_MODAL_ELEMENT` could be an option

dwesolow: updated.

web_ui/src/pages/create-project/components/pose-template/canvas/cursor-context-menu/cursor-context-menu.component.tsx
```diff
@@ -1,65 +1,79 @@
...
+    return createPortal(
+        <div ref={ref} {...overlayProps}>
+            <View
+                position={'absolute'}
+                top={cursorPosition.y}
+                left={cursorPosition.x}
+                zIndex={100001}
+                backgroundColor={'gray-200'}
+                data-testid='position container'
+                {...overlayProps}
```
MarkRedeman, line 72: What if there are multiple modals?

I don't expect this to be a huge deal, since that would be a bad UX. But perhaps we can look into making this more robust in the future?

MarkRedeman (approved):

jpggvilaca (approved):

web_ui/src/pages/create-project/components/pose-template/canvas/cursor-context-menu/cursor-context-menu.component.tsx
dwesolow: I think that's not the case in spectrum right now but agree that we might look into improving this selector.

PR #1147, opened 2025-08-19
Extract edit bounding box (Part 1)

jpggvilaca:
## 📝 Description

* Label components now use hooks directly instead of having context passed as prop
* Refactor Edit bounding box tool (the one to be extracted) by decoupling all dependencies
* Update tests

On part 2 i'll actually move the tool to smart-tools package

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/src/pages/annotator/annotation/labels/shape-label.test.tsx
+++ b/web_ui/src/pages/annotator/annotation/labels/shape-label.test.tsx
@@ -6,10 +6,10 @@ import { screen } from '@testing-library/react';
 import { Annotation } from '../../../../core/annotations/annotation.interface';
 import { ShapeType } from '../../../../core/annotations/shapetype.enum';
 import { SelectedProvider } from '../../../../providers/selected-provider/selected-provider.component';
-import { fakeAnnotationToolContext } from '../../../../test-utils/fake-annotator-context';
 import { getMockedAnnotation } from '../../../../test-utils/mocked-items-factory/mocked-annotations';
 import { getMockedKeypointNode } from '../../../../test-utils/mocked-items-factory/mocked-keypoint';
 import { getMockedLabel } from '../../../../test-utils/mocked-items-factory/mocked-labels';
+import { AnnotationToolProvider } from '../../providers/annotation-tool-provider/annotation-tool-provider.component';
 import { TaskProvider } from '../../providers/task-provider/task-provider.component';
 import { annotatorRender } from '../../test-utils/annotator-render';
 import { ShapeLabel } from './shape-label.component';
@@ -17,11 +17,13 @@ import { ShapeLabel } from './shape-label.component';
 describe('ShapeLabel', () => {
     const renderApp = async (annotation: Annotation) => {
         await annotatorRender(
-            <TaskProvider>
-                <SelectedProvider>
-                    <ShapeLabel annotation={annotation} annotationToolContext={fakeAnnotationToolContext({})} />
-                </SelectedProvider>
-            </TaskProvider>
+            <AnnotationToolProvider>
+                <TaskProvider>
+                    <SelectedProvider>
+                        <ShapeLabel annotation={annotation} />
+                    </SelectedProvider>
+                </TaskProvider>
+            </AnnotationToolProvider>
         );
     };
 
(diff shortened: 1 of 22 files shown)


MarkRedeman:
> On part 2 i'll actually move the tool to smart-tools package

How do we plan to do this if these tools have a hard dependency on hooks like `useTask`, `useAnnotationToolContext` etc.? (in this case any consumer would require to use our `AnnotationToolContextProvider` which has started getting messy and will be hard to use outside of Geti)
What do you think of keeping `annotationToolContext` as a prop that we pass to the tools, but possibly making it simpler: removing parts of its interface that makes it be tight strongly with the Geti codebase?

jpggvilaca:
> > On part 2 i'll actually move the tool to smart-tools package
> 
> How do we plan to do this if these tools have a hard dependency on hooks like `useTask`, `useAnnotationToolContext` etc.? (in this case any consumer would require to use our `AnnotationToolContextProvider` which has started getting messy and will be hard to use outside of Geti) What do you think of keeping `annotationToolContext` as a prop that we pass to the tools, but possibly making it simpler: removing parts of its interface that makes it be tight strongly with the Geti codebase?

This is the PR that takes care of that. I got rid of all dependencies. I can now just copy paste the EditBoundingBox tool. The issue will be the Annotation interface, which i will need to duplicate/simplify. Note that im just moving the EditBoundingBox, the circle/polygon etc will be untouched

web_ui/src/pages/annotator/annotation/labels/shape-label.test.tsx
```diff
@@ -6,22 +6,24 @@ import { screen } from '@testing-library/react';
...
 import { getMockedKeypointNode } from '../../../../test-utils/mocked-items-factory/mocked-keypoint';
 import { getMockedLabel } from '../../../../test-utils/mocked-items-factory/mocked-labels';
+import { AnnotationToolProvider } from '../../providers/annotation-tool-provider/annotation-tool-provider.component';
 import { TaskProvider } from '../../providers/task-provider/task-provider.component';
 import { annotatorRender } from '../../test-utils/annotator-render';
 import { ShapeLabel } from './shape-label.component';
 
 describe('ShapeLabel', () => {
     const renderApp = async (annotation: Annotation) => {
         await annotatorRender(
```
camiloHimura, line 19: Let’s consider adding `AnnotationToolProvider` to the `annotatorRender`

web_ui/src/pages/annotator/tools/edit-tool/edit-annotation-tool.component.tsx
```diff
@@ -32,21 +33,28 @@ const EditAnnotationToolFactory = ({
     disableTranslation = false,
     disablePoints = false,
 }: EditAnnotationToolFactoryProps) => {
-    const { roi } = useROI();
+    const { roi, image } = useROI();
     const { tasks, selectedTask } = useTask();
     const task = selectedTask ?? tasks[0];
+    const { scene } = annotationToolContext;
```
camiloHimura, line 39: Shouldn't we also use `useAnnotationToolContext` here?

jpggvilaca: No need. If i did that then i would need to refactor way more branches than we need. This was the least footprint solution i could do. I did top down without touching almost no code. I added only 3 hooks on the labels component, which were using context from props.

jpggvilaca:
> > On part 2 i'll actually move the tool to smart-tools package
> 
> How do we plan to do this if these tools have a hard dependency on hooks like `useTask`, `useAnnotationToolContext` etc.? (in this case any consumer would require to use our `AnnotationToolContextProvider` which has started getting messy and will be hard to use outside of Geti) What do you think of keeping `annotationToolContext` as a prop that we pass to the tools, but possibly making it simpler: removing parts of its interface that makes it be tight strongly with the Geti codebase?

Discussed offline. This PR solves this, but ofc, there is also the possibility of us duplicating + copy pasting on our side apps, and remove whatever's extraneous.

pplaskie (approved):

camiloHimura (approved):

PR #1177, opened 2025-08-22
Workflow to clean-up old (e.g. > 5 days) daily packages from ghcr

kamillipka:
This PR adds GitHub Actions workflow for cleaning up old packages from daily builds.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


.github/workflows/cleanup-old-packages.yml
```diff
@@ -0,0 +1,113 @@
...
+          # Filter PACKAGE_VERSIONS to contain only daily packages not release packages
+          DAILY_PACKAGE_VERSIONS=$(echo $PACKAGE_VERSIONS | jq '[.[] | select(.tag != null and (.tag | type == "string") and (.tag | test("^[0-9]\\.[0-9]{1,2}\\.[0-9]-[a-z0-9]{8}$"))) | {id, tag, created_at}]')
+
+          NOW=$(date +%s)
+          CUTOFF=$((NOW - MAX_DAYS_OLD*86400))
+
+          # Filter DAILY_PACKAGE_VERSIONS to contain only packages older than MAX_DAYS_OLD
+          OLD_DAILY_PACKAGE_VERSIONS=$(echo "$DAILY_PACKAGE_VERSIONS" | jq --argjson cutoff "$CUTOFF" '[.[] | select((.created_at | strptime("%Y-%m-%dT%H:%M:%SZ") | mktime) < $cutoff)]')
+
+          # Prepare output with packages to be removed
```
AlexanderBarabanov, line 105: In a `dry_run` mode, we may need to view what packages are going to be removed to check that release packages will remain.
Could you please output the list of packages with tags that will be removed?

.github/workflows/cleanup-old-packages.yml
```diff
@@ -0,0 +1,113 @@
...
+#      max_days_old: 5
+#      dry_run: true
+
+name: Cleanup old packages from daily builds
+
+on:
+  workflow_call:
+    inputs:
+      package_name:
+        description: "The name of the package to clean up"
```
AlexanderBarabanov, line 35: In general case we will need to remove several packages - can we use the list as an input?

.github/workflows/cleanup-old-packages.yml
```diff
@@ -0,0 +1,113 @@
...
+# - max_days_old: The maximum age of packages to keep (in days)
+# - dry_run: Perform a dry run without deleting packages
+#
+# Example usage:
+# 1. Call this workflow manually from Github Actions page
+# 2. Trigger this workflow from another workflow using the `workflow_call` event:
+#    uses: ./.github/workflows/cleanup-old-packages.yml
+#    with:
+#      package_name: "your-package-name"
+#      max_days_old: 5
```
mramotowski: As we are going into maintenance mode I assume we will have slower development and we can have lower number of changes on main. Therefore I wouldn't use days but number of builds to keep.

.github/workflows/cleanup-old-packages.yml
```diff
@@ -0,0 +1,113 @@
+# GHCR cleanup workflow
```
mramotowski, line 1: This workflow could be reusable action/workflow in our CI/CD repo and then only called here by some parent workflow if needed.

.github/workflows/cleanup-old-packages.yml
```diff
@@ -0,0 +1,86 @@
...
+          #   {
+          #     "id": package_id,
+          #     "tag": package_first_tag,
+          #     "created_at": date_package_was_pushed_to_ghcr
+          #   }
+          # ]
+          PACKAGE_VERSIONS=$(gh api -H "Accept: application/vnd.github+json" /orgs/open-edge-platform/packages/container/${PACKAGE_NAME//\//%2F}/versions | jq '[.[] | {id, tag: .metadata.container.tags[0], created_at}]')
+
+          # Filter PACKAGE_VERSIONS to contain only daily packages not release packages
+          DAILY_PACKAGE_VERSIONS=$(echo $PACKAGE_VERSIONS | jq '[.[] | select(.tag != null and (.tag | type == "string") and (.tag | test("^[0-9]\\.[0-9]{1,2}\\.[0-9]-[a-z0-9]{8}$"))) | {id, tag, created_at}]')
```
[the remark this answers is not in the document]
mramotowski: I think that it would be better to find all builds that do not work for our scenario that concentrate on daily build only. For example here there are daily builds but there is also something else starting from sha256 that I think we don't need:
image

Consider `^(?!v\d+\.\d+\.\d+(?:-RC\d+)?$).*$` if I'm not wrong it will find everything that is not version that we want to contain. I even added RC there but no sure if it is needed.

.github/workflows/cleanup-old-packages.yml
kamillipka: That was a requirement in ticket,

.github/workflows/cleanup-old-packages.yml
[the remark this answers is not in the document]
kamillipka: I think that it will be better to start with daily builds, then check what are those packages with `sha:<something>` and then create workflow that will delete them

.github/workflows/cleanup-old-packages.yml
kamillipka: Sure, that's why I have implemented `workflow_call`.

.github/workflows/cleanup-old-packages.yml
kamillipka: 1. Shouldn't we create a separate workflow to remove several packages that will call this workflow?
2. `https://github.com/actions/delete-package-versions` - action that is used here to delete packages does not support list of packages as an input, we could use `strategy.matrix` for multiple packages but it would mean that multiple runners will be spawned
3. we can change how we delete packages, instead of using `delete-package-versions` we can use github API call this way everything can be run on one runner and deletion loop will be implemented in shell

AlexanderBarabanov: Thanks for the clarification. As we discussed, let's proceed with the current approach - one package as an input.

mramotowski (approved):
🚢 it!

AlexanderBarabanov (approved):
Thank you!

jcchr (approved):

PR #1187, opened 2025-08-25
Update manifests to support configurable augmentations

kprokofi:
## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [ ] ✅ Tested manually

## ✅ Checklist


libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -45,34 +34,32 @@ class RandomAffine(BaseModelNoExtra):
         description="Whether to apply random affine transformations to the image",
     )
     # Exclude fields as they are supported yet by OTX
-    degrees: float | None = Field(
+    max_rotate_degree: float | None = Field(
```
leoll2, line 67: @maxxgx do these renaming require a migration script?

kprokofi:
merge after: https://github.com/open-edge-platform/training_extensions/pull/4577

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -114,13 +123,120 @@ class GaussianBlur(BaseModelNoExtra):
...
+        default=(0.5, 1.5),
+        title="Saturation range",
+        description="Range of saturation adjustment factors",
+    )
+    hue: tuple[float, float] = Field(
+        default=(-0.05, 0.05),
+        title="Hue range",
+        description="Range of hue adjustment values",
+    )
+    p: float = Field(
```
maxxgx: I think `probability` would be preferable over `p` or `prob`

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -114,13 +123,120 @@ class GaussianBlur(BaseModelNoExtra):
...
+    enable: bool = Field(
+        default=True,
+        title="Enable photometric distort",
+        description="Whether to apply photometric distortion to the image",
+    )
+    brightness_delta: int = Field(
+        default=32,
+        ge=0,
+        title="Brightness delta",
+        description="Maximum delta for brightness adjustment",
```
maxxgx: Could also describe the value's unit? For example, what does `brightness_delta=32` mean? Does upper limit applies here?

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -114,13 +123,120 @@ class GaussianBlur(BaseModelNoExtra):
...
+    )
+    contrast: tuple[float, float] = Field(
+        default=(0.5, 1.5),
```
maxxgx: Why `1.5`?

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -114,13 +123,120 @@ class GaussianBlur(BaseModelNoExtra):
...
+        title="Contrast range",
+        description="Range of contrast adjustment factors",
+    )
+    saturation: tuple[float, float] = Field(
+        default=(0.5, 1.5),
+        title="Saturation range",
+        description="Range of saturation adjustment factors",
+    )
+    hue_delta: int = Field(
```
maxxgx: As above, can you add the unit and range if applicable?

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -45,34 +34,32 @@ class RandomAffine(BaseModelNoExtra):
...
-        exclude=True,
     )
-    translate_y: float | None = Field(
-        default=None,
-        ge=0.0,
-        lt=1.0,
-        title="Vertical translation",
-        description="Maximum vertical translation as a fraction of image height",
-        exclude=True,
+    scaling_ratio_range: tuple[float, float] | None = Field(
```
maxxgx: I think it's better to split this into 2 parameters: `scaling_ratio_min` and `scaling_ratio_max`. Then you can add limits individually to the fields (ge=0.0, le=1.0). Additionally, you can add a model validator:
```python
@model_validator(mode="after")
    def validate_scaling_ratio_range(self) -> "RandomAffine":
        if self.scaling_ratio_min > self.scaling_ratio_max:
            raise ValueError("Incorrect range, lower bound scaling ratio cannot be higher than the upper bound")
        return self
```

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
maxxgx: Since you are adding the support for these parameters, you can delete the comment at L36. I see you added actual default value, in this case you can make this parameter required. 

Note: you probably need to update the unit tests for the rest views in the director service since now these parameters will be exposed as well

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
kprokofi: It is default value from torchvision augmentation

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
kprokofi: Unfortunately, we should be consistent with naming in OTX. It is the easiest way to provide this functionality.

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
kprokofi: Done

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
kprokofi: Done

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
maxxgx: I don't think that's a sufficient reason to allow inconsistent naming. Since `p` and `prob` will be exposed via the REST API as keys for these parameters, it's important to maintain consistency. Once the `config.yaml` is loaded in OTX, you can convert all `probability` keys to either p or prob. The presentation layer should not be affected by implementation details like this.

interactive_ai/services/auto_train/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM python:3.10-slim-bookworm@sha256:66a6e8a211a02794515d4e1aa1970fc037ec4dde14
 FROM base AS build
 
 # Install Git and clone the OTX at the specified tag
-ARG OTX_RELEASE_TAG=release/2.5
+ARG OTX_RELEASE_TAG=develop
```
maxxgx, line 6: Are we sure we want to target the `develop` branch? It seems to very risky considering that its structure could change significantly due to ongoing development. Could another release be made instead?

CC: @leoll2

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -6,35 +6,31 @@
...
+        "Note: this augmentation is not supported when Tiling algorithm is enabled.",
+    )
+    crop_ratio_range: list[float] | None = Field(
+        default=None,
+        title="Crop resize ratio range",
+        description=(
+            "Range (min, max) of crop ratios to apply during resize crop operation. "
+            "Specifies the fraction of the original image dimensions to retain after cropping. "
+            "For example, (0.8, 1.0) will randomly crop between 80% and 100% of the original size. "
+            "Both values should be between 0.0 and 1.0."
```
maxxgx, line 23: You can enforce `"Both values should be between 0.0 and 1.0."` with pydantic:
```python
    @model_validator(mode="after")
    def validate_crop_range(self) -> "RandomResizeCrop":
        if self.crop_ratio_range is None:
            return self
        if len(self.crop_ratio_range) != 2:
            raise ValueError("crop_ratio_range must be a list of exactly two float values")
        if self.crop_ratio_range[0] >= self.crop_ratio_range[1]:
            raise ValueError("The first value in crop_ratio_range must be less than the second value")
        if not (0 <= self.crop_ratio_range[0] <= 1) or not (0 <= self.crop_ratio_range[1] <= 1):
            raise ValueError("Values in crop_ratio_range must be between 0 and 1")
        return self
```

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
```diff
@@ -6,35 +6,31 @@
...
+        ),
     )
-    # Exclude fields as they are supported yet by OTX
-    ratio: float | None = Field(
-        gt=0.0,
+    aspect_ratio_range: list[float] | None = Field(
```
maxxgx, line 26: Is there a reason why you replaced `tuple` with `list`?

List fields with no extra validation allow for arbitrary number of items in the list. User can, through rest API, submit and save invalid configuration such as `aspect_ratio_range=[0.0, 2.3, -0.5, -10]`. 

If the range value is restricted (e.g. [0, 1]), please add validation for that as well to avoid potential errors during training.

Same applies to other `list` type fields.

interactive_ai/services/auto_train/Dockerfile
leoll2: While developing new features, yes we target the `develop` branch. Before every Geti releases, OTX will also produce a release candidate that can be used as reference from the Geti release branch. Example: Geti's `release-2.13` could target OTX `release/2.6`.

kprokofi: For now we didn't create a release branch, but develop includes all latest results. I could switch it later to the new release branch. I hope it safe since we don't touch algorithmic part right now.

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
kprokofi: Done https://github.com/open-edge-platform/geti/pull/1187/commits/85b392d844dc833c0874f35c0c6db648304cf2f3

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
kprokofi: Done https://github.com/open-edge-platform/geti/pull/1187/commits/85b392d844dc833c0874f35c0c6db648304cf2f3

libs/configuration_tools/src/geti_configuration_tools/hyperparameters/augmentation.py
maxxgx: It's not required because we previously only allowed for `enable` on/off, other fields such as `degrees` were excluded from serialization and could not be set by the user

maxxgx (approved):

interactive_ai/supported_models/geti_supported_models/parser.py
```diff
@@ -25,7 +25,7 @@ def parse_manifest(*manifest_sources, relative: bool = True) -> ModelManifest:
         manifest_sources = tuple(str(resources.files(manifests).joinpath(path)) for path in manifest_sources)
     yaml_manifest = hiyapyco.load(
         *manifest_sources,
-        method=hiyapyco.METHOD_MERGE,
+        method=hiyapyco.METHOD_SUBSTITUTE,
```
leoll2, line 28: Can you explain this change?

interactive_ai/supported_models/geti_supported_models/manifests/classification/base.yaml
```diff
@@ -3,18 +3,55 @@ task: classification
 hyperparameters:
   dataset_preparation:
     augmentation:
+      random_resize_crop:
+        enable: true
+        crop_ratio_range:
+          - 0.08
+          - 1.0
```
leoll2, line 10: If this parameter represents a range, why not represent the two values as "min" and "max"?

leoll2 (approved):

interactive_ai/supported_models/geti_supported_models/parser.py
kprokofi: We discussed this with Max, this change is needed to properly propagate all the values from manifests. If we will not use Substitute method, all values will be just merged. For example, "crop ratio range" from specific model manifest will be merged with "crop ratio range" from base.yaml. It causes problems.

interactive_ai/supported_models/geti_supported_models/manifests/classification/base.yaml
kprokofi: Because it is much simpler to pass the whole range right away directly to augmentation in OTX. This implementation is not ideal, but at least we avoid a lot of mapping on training backend side.

PR #1216, opened 2025-09-01
Simplify zoom logic to avoid react 19 regressions

MarkRedeman:
## 📝 Description

This PR simplifies our logic for zooming into an image and/or annotation (when in the sub task of a task chain project). It does this by first removing state from the zoom provider that can be computed based on other values of its state. We also no longer keep track of the `zoomState` in the `ZoomProvider`, instead we use a new `ZoomStateProvider` that takes this value directly from the `react-zoom-pan-pinch` package, which allows us to remove `SyncZoomState`.
Later we can choose to remove `zoomState` from `ZoomContextProps` and instead return this from a dedicated hook. This will improve some of the performance of ourcomponentss as they will have less reasons to rerender.
Lastly the logic for updating the zoom target was duplicated by `SyncZoomTarget`: the same logic is included in `useZoomIntoAnnotation`, so the component is now removed.

- Remove more unnecessary components
- Make initialZoomState a computed value
- Remove unnecessary components changing the zoom target
  This logic is fully handled (and was thus duplicated) via the
  `useZoomIntoAnnotation` hook.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist


web_ui/src/pages/annotator/zoom/sync-zoom-state.component.tsx
```diff
@@ -53,61 +13,5 @@ export const SyncZoomTarget = () => {
     - Zoom in on the annotation if the selectedTask is a local task
 */
 export const SyncZoomState = () => {
```
dwesolow: leftovers ;)

web_ui/src/pages/annotator/zoom/use-zoom-into-annotations.hook.ts
```diff
@@ -25,7 +25,7 @@ export const useZoomIntoAnnotation = (): void => {
         [taskAnnotations]
     );
 
-    const zoomIntoAnnotation = useCallback(() => {
+    const zoomIntoAnnotation = useCallback((): ZoomTarget => {
```
dwesolow: this function no longer zooms into annotation, it returns zoom target, could we rename it?

web_ui/dev-proxy.ts
```diff
@@ -66,13 +66,19 @@ export const devProxy: ProxyConfig = [
                 return buffer;
             }
 
-            const originalFeatureFlags = JSON.parse(buffer.toString('utf8'));
+            const content = buffer.toString('utf8');
```
jpggvilaca: why this change (the whole file)? If needed, could we make it on a different PR?, seems a bit out of scope of this one :P

web_ui/src/pages/annotator/zoom/sync-zoom-state.component.tsx
```diff
@@ -1,47 +1,7 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-import { useEffect } from 'react';
```
jpggvilaca, line 4: why not delete the whole file?

web_ui/src/pages/annotator/zoom/use-zoom-into-annotations.hook.ts
```diff
@@ -54,23 +50,19 @@ export const useZoomIntoAnnotation = (): void => {
         // We need to compare the image target (which is the initial target) with the current zoomTarget
         const imageTargetConfig = { x: 0, y: 0, width, height } as ZoomTarget;
 
-        const targetChanged = !isEqual(imageTargetConfig, zoomTarget);
-
-        if (targetChanged) {
-            setZoomTarget(imageTargetConfig);
-        }
+        return imageTargetConfig;
```
jpggvilaca, line 53: feels weird that "resetZoom" returns anything. If we really need this shall we rename this to something like `originalZoom`?

web_ui/dev-proxy.ts
MarkRedeman: Moved to #1217

web_ui/src/pages/annotator/zoom/use-zoom-into-annotations.hook.ts
MarkRedeman: Yep, renamed it to `getSelectedAnnotationZoomTarget`.
I think overall this hook could be simplified a bit more as well but didn't want to bother with that too much atm.

web_ui/src/pages/annotator/zoom/zoom-provider.component.tsx
```diff
@@ -153,64 +144,58 @@ export const ZoomProvider = ({ children }: ZoomProviderProps) => {
...
-        zoomState,
-        setZoomState,
-
-        isPanning,
-        isPanningDisabled,
-        setIsPanningDisabled,
-
-        isDblCLickDisabled,
-        setIsDblClickDisabled,
+        zoomState: defaultZoomState,
```
dwesolow, line 147: Could we remove this since we have separate provider for the zoom state?

web_ui/src/pages/annotator/zoom/zoom-provider.component.tsx
```diff
@@ -76,75 +58,84 @@ interface ResetInitialZoomProps {
...
+
+    const zoomState = {
+        zoom: transformState.scale,
+        translation: {
+            x: transformState.positionX,
+            y: transformState.positionY,
+        },
+    };
+
+    return <ZoomStateContext.Provider value={zoomState}>{children}</ZoomStateContext.Provider>;
```
dwesolow, line 114: Why do we need extra provider for the zoom state?

web_ui/src/pages/annotator/zoom/zoom-provider.component.tsx
MarkRedeman: I would like to do that in a follow up PR. If we would remove it now then we need to update all components that call `useZoom` and use `zoomState`, which is quite a lot.

web_ui/src/pages/annotator/zoom/zoom-provider.component.tsx
MarkRedeman: This component had to be rendered after `TransformWrapper` because otherwise `useTransformEffect` and `useTransformContext` are not available.

We keep this value in the context so that we only need to have 1 `useTransformEffect` (which I believe is just a simple `useEffect`) for all of the components that use the zoom state.
We have roughly 15 components that rely on `zoomState` most of which are rendered multiple times (e.g. the annotations) so now these all will be able to use the same state vs keeping their own.

dwesolow (dismissed):
Thanks!

dwesolow (approved):

jpggvilaca (approved):

PR #1235, opened 2025-09-02
Update smart-tools dependency versions

jpggvilaca:
## 📝 Description

* When installing packages on geti tune, the latest version of onnx was being installed, which isnt backwards compatible (we use 1.16 on geti classic), so now i updated `^` to `~` to make sure we only update minor versions.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/web_ui/.gitignore
+++ b/web_ui/.gitignore
@@ -39,3 +39,5 @@ tests/e2e/.auth
 
 public/openapi-spec.json
 .nx/
+
+/public/onnxruntime-web/
--- a/web_ui/.gitignore
+++ b/web_ui/.gitignore
@@ -39,5 +39,3 @@ tests/e2e/.auth
 
 public/openapi-spec.json
 .nx/
-
-/public/onnxruntime-web/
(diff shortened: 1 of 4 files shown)


MarkRedeman (dismissed):
I'd prefer to dive a bit more into why this happens. We should have the same pacakges structure in Geti Tune as we have in this repo so I don't fully understand why this setup raises issues.

I can see how this copying will work though, however one issue it has is that it will prevent us from applying cache busting, so I'd prefer the previous approach.

jpggvilaca:
> I'd prefer to dive a bit more into why this happens. We should have the same pacakges structure in Geti Tune as we have in this repo so I don't fully understand why this setup raises issues.
> 
> I can see how this copying will work though, however one issue it has is that it will prevent us from applying cache busting, so I'd prefer the previous approach.

Added explanation

MarkRedeman (approved):

dwesolow (approved):

PR #1247, opened 2025-09-04
GH 1228 fixed charts location

piotrgrubicki:
This PR rearranges the location of independent helm charts so that they are picked up by the makefiles and published.


mpryahin (approved):
Savage work here, thanks Piotr!

jcchr (approved):

PR #1287, opened 2025-09-09
ffmpeg upgrade

jcchr:
## 📝 Description

As the ffmpeg in the 5.1.6-0+deb12u1 version has been removed from the repo - we need to update its version. This is the goal of this PR.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Check if after changing the ffmpeg version - projects can be handled (images are handled correctly, training/inference works, etc)

- [ ] ✅ Tested manually

## ✅ Checklist


interactive_ai/migration_job/Dockerfile
```diff
@@ -27,7 +27,7 @@ FROM base AS runtime
 
 # Install ffmpeg and curl
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends ffmpeg=7:5.1.6-0+deb12u1 curl && \
+    apt-get install -y --no-install-recommends ffmpeg=7:5.1.7-0+deb12u1 curl && \
```
leoll2:
```suggestion
    apt-get install -y --no-install-recommends ffmpeg=7:5.1.* curl && \
```
In one month you'll have the same problem when the deb update is superseded by the new one.

jcchr: I checked - and last change of the version of ffmpeg was made more than 5 month ago. Thus I would rather stick to an exact version - to avoid potential problems with backward compatibility of newer versions - having in mind, that from time to time we will have to refresh ffmpeg version.

AlbertvanHouten: I am sure ffmpeg follows a normal release versioning process and will not introduce breaking changes in a patch version release. It does not make sense to break Geti multiple times a year by setting an exact version that can disappear at any moment. For a lot of our python dependencies, we already allow flexible patch versions  and we should do the same for ffmpeg.

AlbertvanHouten (approved):

leoll2 (approved):

PR #1254, opened 2025-09-05
Fix wrong frame being returned for VFR video

AlbertvanHouten:
Fixes: https://github.com/open-edge-platform/geti/issues/1246

## 📝 Description
This pull request improves the accuracy of video frame extraction by introducing support for millisecond-based seeking using the video's frames per second (fps) value. This is necessary to account for videos with a variable frame rate and is using the same logic as the inference service returning the predictions. 

**Video frame extraction accuracy improvements:**

* Updated the `decode` method in `_VideoDecoderInterface` and `_VideoDecoderOpenCV` to accept an optional `fps` parameter, and use millisecond-based seeking (`cv2.CAP_PROP_POS_MSEC`) when fps is provided, falling back to frame-based seeking otherwise. [[1]](diffhunk://#diff-5288dacce38c36e7547b24bf22c09902be075d3676ccee431d53085f5a320fedL162-R162) [[2]](diffhunk://#diff-5288dacce38c36e7547b24bf22c09902be075d3676ccee431d53085f5a320fedL226-R233) [[3]](diffhunk://#diff-5288dacce38c36e7547b24bf22c09902be075d3676ccee431d53085f5a320fedL250-R265)
* Modified `VideoFrameReader.get_frame_numpy` to accept and pass the optional `fps` parameter to the decoder, and updated its usage throughout the codebase. [[1]](diffhunk://#diff-21b0de0193d4e2b66f52a4b6358381d81d3d5935f0489b96b37a8d16f9997d99L22-R31) [[2]](diffhunk://#diff-21b0de0193d4e2b66f52a4b6358381d81d3d5935f0489b96b37a8d16f9997d99L37-R41) [[3]](diffhunk://#diff-3e10a4c0836ce2c40b546be788021bc92a74a2a953a29db46ff8b4cecd7afe79R74)

**Unit test enhancements:**

[...]

PR #1293, opened 2025-09-10
Fix variable framerate video issues 

### LINKED ISSUE #1246: Predictions for variable framerate video's are out of sync
### 🧩 Issue type

Bug 🐞

### 📄 Description

Predictions on video's with a variable framerate can be out of sync in certain cases when requesting single frame inference. When using the batch predict endpoint, the predictions are lining up correctly. 

### 📋 Steps to reproduce (if applicable)

See for example this frame: https://10.91.242.204/organizations/6b076e21-5e87-4f00-b3c8-bd87f8549fe7/workspaces/9462bdee-b900-4cf7-92e9-4b8e7821b368/projects/55061e18619099bee4239554/datasets/55061e18619099bee423956b/annotator/video/55062053619099bee4239600/700

Also note that if playing the video with predictions, the correct predictions are shown. 

### 💻 Expected behavior

Single frame predictions line up with the requested frame from the user. 

### 🖼️ Screenshots, video recordings or logs

Image

### 📦 Geti version

2.12

### Hardware

_No response_

### 🖥️ Operating System

_No response_

### Browser (if applicable)

_No response_

### Python Version (if relevant)

_No response_

### ➕ Additional context

_No response_

AlbertvanHouten:
## 📝 Description
This pull request adds a check for uploaded videos and ensure that every video has a constant frame rate. If a variable frame rate video is uploaded, it is converted to have a constant frame rate. 

It also introduces a new data migration to convert variable frame rate (VFR) videos to constant frame rate (CFR) within the video storage system. The migration ensures video frame rates are standardized, which helps maintain annotation alignment and improves downstream processing. This migration script is disabled by default through a feature flag. Right now, it's only planned for 1 customer. 

This PR also reverts the previously updated video decoding logic from https://github.com/open-edge-platform/geti/pull/1254 as this will mess with already annotated VFR videos from previous versions. 

**Migration for VFR to CFR video conversion:**

* Added a migration script `convert_vfr_videos_to_cfr.py` that detects VFR videos, extracts all frames, reconstructs them at a constant frame rate, and replaces the original video in storage. The script uses OpenCV and ffmpeg for processing and is integrated with Minio and MongoDB.
* Added metadata for the migration, specifying its description, affected collections, and binary data impact.
* Added a changelog entry for the new migration script and its metadata.

**Versioning:**

* Bumped the data version from 34.0 to 35.0 to reflect the new migration.

image

Resolves https://github.com/open-edge-platform/geti/issues/1246

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually


interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
```diff
@@ -0,0 +1,392 @@
...
+
+from migration.utils import IMigrationScript, MongoDBConnection
+from migration.utils.connection import MinioStorageClient
+
+logger = logging.getLogger(__name__)
+
+BUCKET_NAME_VIDEOS = os.environ.get("BUCKET_NAME_VIDEOS", "videos")
+
+
+class FeatureFlagProvider:
```
leoll2: If you can't reuse the implementation in `feature_tools`, I suggest to move this class to `migration.utils`

interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
```diff
@@ -0,0 +1,392 @@
...
+
+                    # Check if video is VFR
+                    if cls._is_variable_frame_rate(url):
+                        logger.info(f"Converting VFR video {video_id} to CFR")
+
+                        # Get the FPS for reconstruction
+                        fps = cls._get_video_fps(url)
+
+                        # Convert VFR to CFR
+                        converted_file = cls._convert_vfr_to_cfr(url, fps)
```
leoll2, line 66: This operation may fail if there is not enough disk space to store the converted file. I don't think there is a reasonable way to handle such an exception at runtime, but please inform the customer that he should have enough disk space before attempting the upgrade.

interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
```diff
@@ -0,0 +1,392 @@
...
+            [  # noqa: S607
+                "ffprobe",
+                "-v",
+                "0",
+                "-select_streams",
+                "v:0",
+                "-show_entries",
+                "stream=r_frame_rate",
+                "-of",
+                "csv=p=0",
```
leoll2, line 135: Any reason for using `csv` here and `json` at L134? Are both returning the same r_frame_rate?

interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
```diff
@@ -0,0 +1,392 @@
...
+            ],
+            capture_output=True,
+            check=False,
+            text=True,
+        )
+
+        # Get avg_frame_rate
+        avg_frame_rate_result = subprocess.run(  # noqa: S603
+            [  # noqa: S607
+                "ffprobe",
```
leoll2, line 146: Is it possible to extract both r_frame_rate and avg_frame_rate with a single ffprobe call? Just wondering since ffprobe takes some time to run, especially for long videos.

interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
```diff
@@ -0,0 +1,392 @@
...
+                try:
+                    # Get presigned URL for the video
+                    url = storage_client.presigned_get_object(BUCKET_NAME_VIDEOS, video_path)
+
+                    # Check if video is VFR
+                    if cls._is_variable_frame_rate(url):
+                        logger.info(f"Converting VFR video {video_id} to CFR")
+
+                        # Get the FPS for reconstruction
+                        fps = cls._get_video_fps(url)
```
leoll2, line 63: The function returns the r_frame_rate (raw frame rate) which, based on my understanding, is the value reported by the video container. For VFR videos, however, the frame rate changes throughout the video; I'm wondering, is this the right fps value to use for reconstruction? Why not use avg_frame_rate to ensure that every frame in the original video is mapped to a frame in the converted one? The 1:1 correspondence would then guarantee that the annotations (which are tied to the frame index) stay valid after the conversion.

interactive_ai/services/resource/app/usecases/media_uploaded_usecase.py
```diff
@@ -121,6 +125,58 @@ def crop_to_thumbnail(pil_image: PILImage.Image, target_height: int, target_widt
         y2 = round(min(y2, resized_image.height))
         return resized_image.crop((x1, y1, x2, y2))
 
+    @staticmethod
+    def _convert_vfr_to_cfr_if_needed(
```
leoll2: Rename suggestion: `ensure_constant_frame_rate`

interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
AlbertvanHouten: The core information is still the same, just presented in a different format. This function was copied from the video decoder and the others were newly implemented, hence the difference.

interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
AlbertvanHouten: IMO, it makes more sense to map to the value that the video container was originally meant to be, which is the r_frame_rate. VFR video's will end up with a higher or lower average frame rate, but either way, both will still maintain the annotation to frame index mapping. An annotated frame will still have the same index in the newly reconstructed video, only the timestamp will be different. 

The timestamp is used in go services for prediction, but this extracts the fps value in the same way as is done here to locate where the frame is.

interactive_ai/data_migration/migration/scripts/convert_vfr_videos_to_cfr.py
AlbertvanHouten: This command is near instant. It's already used for every single frame inference request within the inference gateway.

leoll2 (approved):

jcchr (approved):

PR #1348, opened 2025-09-30
ci: Add workflow for builder images management

mramotowski:
## 📝 Description

Add workflow for builder images management. They will be published in ghcr.io/open-edge-platform/geti/builders/IMAGE_NAME registries. 

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/.github/workflows/builder-images.yml
+++ b/.github/workflows/builder-images.yml
@@ -1,0 +1,105 @@
+# This workflow builds and publishes builder container images for Python and Go.
+# These images contain all the tools required to build applications and are used
+# both in GitHub Actions workflows and for local development. The workflow checks
+# for changes in the builder image directories, builds the relevant images if needed,
+# and pushes them to the GitHub Container Registry. This ensures that up-to-date
+# builder images are available for creating product images in CI/CD pipelines and
+# for developers working locally.
+#
+# NOTICE:
+# Whenever you make changes to the builder image directories,
+# you MUST bump the BUILDER_TAG in the corresponding Makefile.
+# This ensures that the new image is tagged and published correctly.
+
+name: Build and Publish Builder Images
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'dev_tools/builder_images/**'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'dev_tools/builder_images/**'
+
+permissions: {}
+
+jobs:
+  check_paths:
+    name: Check if workflow should run
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      build_list: ${{ steps.changed-files-yaml.outputs.modified_keys }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+
+      - name: Get all paths that should trigger the workflow
+        id: changed-files-yaml
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
+        with:
+          files_yaml: |
+            go:
+              - dev_tools/builder_images/go-builder/**
+            python:
+              - dev_tools/builder_images/python-builder/**
+  
+  create_builder:
+    name: Create and push builder images
+    needs: check_paths
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write # To push to GitHub Container Registry
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - path: dev_tools/builder_images/go-builder
+            run_condition: ${{ contains(needs.check_paths.outputs.build_list, 'go')}}
+          - path: dev_tools/builder_images/python-builder
+            run_condition: ${{ contains(needs.check_paths.outputs.build_list, 'python')}}
+    env:
+      REGISTRY: ghcr.io/open-edge-platform
+    steps:
+      - name: Checkout repository
+        if: ${{ matrix.run_condition }}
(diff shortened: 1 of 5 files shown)


.github/workflows/builder-images.yml
```diff
@@ -0,0 +1,80 @@
+name: Build and Publish Builder Images
```
AlexanderBarabanov, line 14: Please add a comment with a brief description of this workflow.

.github/workflows/builder-images.yml
```diff
@@ -0,0 +1,80 @@
...
+      - main
+    paths:
+      - 'dev_tools/builder_images/**'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'dev_tools/builder_images/**'
+
+permissions:
```
AlexanderBarabanov: For a consistency, please move this permission to the job that requires it - `check_paths` and define `permissions: {}` on the workflow level

.github/workflows/builder-images.yml
```diff
@@ -0,0 +1,80 @@
...
+            python:
+              - dev_tools/builder_images/python-builder/**
+  
+  create_builder:
+    name: Create and push builder images
+    needs: check_paths
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
AlexanderBarabanov: Please add a short comment in the YAML to explain why this job requires `packages: write` - this will help to maintain this workflow in future.

.github/workflows/builder-images.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+
+      - name: Get all paths that should trigger the workflow
+        id: changed-files-yaml
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
+        with:
+          files_yaml: |
```
AlexanderBarabanov, line 48: If PR changes only `dev_tools/builder_images/Makefile.shared-builders`  and does not touch `go-builder` or `python-builder` folders - shall we also execute `create_builder` and try to build all images?

.github/workflows/builder-images.yml
mramotowski: I think name here is self explanatory I can copy paste name in comment and add detection like this: `# Build and Publish Builder Images` but I don't know if it makes sense here as this don't have any inputs variables.

.github/workflows/builder-images.yml
mramotowski: I think we want to build images only when Dockerfile or version in corresponding builder image Makefile change. What do you think?

.github/workflows/builder-images.yml
AlexanderBarabanov: I think we can briefly explain what builder images are and where we use them.

``
This workflow builds and publishes container images (builder images) that are used to create product images (Python and Go) in both GitHub Actions workflows and local development, images contains all the tools required to build applications
``

.github/workflows/builder-images.yml
AlexanderBarabanov: Yes, agree. As this file is not required for container image building step, we may exclude it.

AlexanderBarabanov (commented):
Thanks! Some opens that I think we should cover in this PR:

1) As far as I understand, these builder images and files are used in Main CI, so while making any PR into `dev_tools/builder_images` folder we may need to ensure that this PR (especially new builder images that we will be pushed as a result of this PR) will not break Main CI - can we add some checks in `builder-images.yml` to cover it?

2) PR into Dockerfile in `dev_tools/builder_images` folder will produce and push a new version of builder images - do we also need to change corresponding Makefile and update `BUILDER_TAG` ? If yes - if we need to do it manually - I propose to add check in the `builder-images.yml` workflow that will verify that PR contains a new BUILDER_TAG value as well - if not, workflow need to fail and inform contributor, that a Makefile must also be changed with a new `BUILDER_TAG`.

mramotowski:
> 1. As far as I understand, these builder images and files are used in Main CI, so while making any PR into `dev_tools/builder_images` folder we may need to ensure that this PR (especially new builder images that we will be pushed as a result of this PR) will not break Main CI - can we add some checks in `builder-images.yml` to cover it?

It is not needed as this is 2 steps change. First you need to update images as one PR. Then you need to change a Main CI as second PR it won't work in other case.


> 2\. PR into Dockerfile in `dev_tools/builder_images` folder will produce and push a new version of builder images - do we also need to change corresponding Makefile and update `BUILDER_TAG` ? If yes - if we need to do it manually - I propose to add check in the `builder-images.yml` workflow that will verify that PR contains a new BUILDER_TAG value as well - if not, workflow need to fail and inform contributor, that a Makefile must also be changed with a new `BUILDER_TAG`.

Good point. I will add it.

kamillipka (dismissed):
LGTM, thanks.

AlexanderBarabanov (commented):
In https://github.com/open-edge-platform/geti/blob/main/.github/workflows/component.yml#L53-L56 we have:
```
      REGISTRY: ${{ secrets.REGISTRY }}
      PYTHON_BUILDER_IMAGE: builder-images/python-builder:v0.1
      GO_BUILDER_IMAGE: builder-images/go-builder:v0.1
```
Does PR chat update Dockerfile also needs to change this versions in a `component.yml` as well (in this case I think we may face with race condition)? Or shall it be done in a second step?

As this process critical for CI stability and looks like requires several steps – I would recommend to create a Readme file  in `dev_tools/builder_images` folder with a description of what needs to be done to properly upgrade Dockerfile.

AlexanderBarabanov (approved):

jcchr (approved):

PR #1417, opened 2025-10-22
remove cuda components with wrong licenses

jcchr:
## 📝 Description

PR removes from training images nvidia libraries not covered by EULA license.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Check if all models available in Geti can be trained using new images.

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/interactive_ai/workflows/train/trainer/gpu/Dockerfile
+++ b/interactive_ai/workflows/train/trainer/gpu/Dockerfile
@@ -100,7 +100,7 @@ RUN rm -rf /interactive_ai/workflows/train/trainer/.venv/lib/python3.10/site-pac
     /interactive_ai/workflows/train/trainer/.venv/lib/python3.10/site-packages/nvidia/cuda_nvrtc/lib/libnvrtc.alt.so.12 \
     /interactive_ai/workflows/train/trainer/.venv/lib/python3.10/site-packages/nvidia/cuda_nvrtc/lib/libnvrtc-builtins.alt.so.12.8
 
-FROM base AS runtime
+FROM python:3.10-slim-bookworm@sha256:cda0e2fa3894f9ada2c652534e6de6b2fe65687c63208a77e9e8d88231547902 AS runtime
 
 RUN ulimit -n 65536
 
--- a/interactive_ai/workflows/train/trainer/gpu/Dockerfile
+++ b/interactive_ai/workflows/train/trainer/gpu/Dockerfile
@@ -16,23 +16,33 @@ ENV TORCH_CUDA_ARCH_LIST="6.0;6.1;6.2;7.0;7.2;7.5;8.0;8.6"
 ENV TORCH_DONT_CHECK_COMPILER_ABI=1
 
 RUN apt update \
-	&& apt install -y wget --no-install-recommends
+        && apt install -y wget --no-install-recommends
 
 RUN wget https://developer.download.nvidia.com/compute/cuda/repos/debian12/x86_64/cuda-keyring_1.1-1_all.deb \
     && dpkg -i cuda-keyring_1.1-1_all.deb \
     && rm -rf cuda-keyring_1.1-1_all.deb
 
 RUN apt update \
-	&& apt install -y --no-install-recommends \
+        && apt install -y --no-install-recommends \
     cuda-libraries-12-9 \
     cuda-compiler-12-9 \
     && apt clean all \
     && rm -rf /var/cache/apt/* \
-	&& echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
-	&& echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+        && echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
+        && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+
+RUN rm -rf /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc.alt.so.12 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc.alt.so.12.9.86 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc-builtins.alt.so.12.9 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libcusolverMg.so.11 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc-builtins.alt.so.12.9.86 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libcusolverMg.so.11.7.5.82 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libcufilt.a
 
 # step 2: get python dependencies
-FROM base AS runtime
+FROM base AS python_dependencies
+
+RUN ulimit -n 65536
 
 ENV UV_COMPILE_BYTECODE=1
 ENV UV_LINK_MODE=copy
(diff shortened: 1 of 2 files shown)


interactive_ai/workflows/train/trainer/gpu/Dockerfile
```diff
@@ -16,23 +16,33 @@ ENV TORCH_CUDA_ARCH_LIST="6.0;6.1;6.2;7.0;7.2;7.5;8.0;8.6"
...
-	&& apt install -y --no-install-recommends \
+        && apt install -y --no-install-recommends \
     cuda-libraries-12-9 \
     cuda-compiler-12-9 \
     && apt clean all \
     && rm -rf /var/cache/apt/* \
-	&& echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
-	&& echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+        && echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
+        && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
```
leoll2: The identation seems off here. Also at L19 and L26.

interactive_ai/workflows/train/trainer/gpu/Dockerfile
```diff
@@ -16,23 +16,33 @@ ENV TORCH_CUDA_ARCH_LIST="6.0;6.1;6.2;7.0;7.2;7.5;8.0;8.6"
...
+
+RUN rm -rf /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc.alt.so.12 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc.alt.so.12.9.86 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc-builtins.alt.so.12.9 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libcusolverMg.so.11 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc-builtins.alt.so.12.9.86 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libcusolverMg.so.11.7.5.82 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libcufilt.a
```
leoll2, line 40: I am a bit afraid of removing so many library files in such a manual fashion. The risk is not only to break the training of some model architectures, but also to silently slow it down. I wouldn't proceed with this change unless you have extensive benchmarks to exclude any functional/performance regression.

Could you also elaborate on why these files need to be removed in the first place? When you install `cuda-libraries` and `cuda-compiler`, you accept the license they're distributed with. I don't understand how removing some files would make any difference in that respect.

@open-edge-platform/training_extensions-lib-write please review these changes.

interactive_ai/workflows/train/trainer/xpu/Dockerfile
```diff
@@ -2,6 +2,25 @@
...
+COPY --link uv.lock uv.lock
+COPY --link pyproject.toml pyproject.toml
+
+# For OTX to install torch with ipex support, else defaults to cpu version
+RUN --mount=type=cache,target=/root/.cache/uv uv sync --frozen --no-dev --no-editable
+RUN uv pip uninstall torch torchvision
+RUN uv pip install torch==2.8.0 torchvision==0.23.0 --default-index https://download.pytorch.org/whl/xpu
+
+RUN rm -rf /interactive_ai/workflows/train/trainer/.venv/lib/python3.10/site-packages/nvidia \
+           /interactive_ai/workflows/train/trainer/.venv/lib/python3.10/site-packages/triton/backends/nvidia/
```
leoll2, line 22: Triton has MIT license. What's the problem with it?

interactive_ai/workflows/train/trainer/gpu/Dockerfile
jcchr: We need to remove those file - as they cannot be redistributed by us. List of files that can be redistributed is stated here - https://docs.nvidia.com/cuda/eula/index.html#cuda-toolkit-supplement-license-agreement. So you can install them locally on your own - but you cannot distribute them.
As the rest of files which are a part of nvidia drivers/libraries needs to stay - the only option was to remove only certain files - those that were not allowed to be redistributed.
I agree - that this approach adds a risk of errors - that is why I tested it extensively on all models we have in Geti. Additionally I checked the purpose of removed files - and they shouldn't be used by us.

interactive_ai/workflows/train/trainer/xpu/Dockerfile
jcchr: Inside of this folder there were libraries from nvidia that shouldn't be redistributed by us - and as we don't use triton - I decided to get rid of the whole folder.

interactive_ai/workflows/train/trainer/gpu/Dockerfile
```diff
@@ -16,23 +16,33 @@ ENV TORCH_CUDA_ARCH_LIST="6.0;6.1;6.2;7.0;7.2;7.5;8.0;8.6"
...
     && apt clean all \
     && rm -rf /var/cache/apt/* \
-	&& echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
-	&& echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+    && echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
+    && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+
+RUN rm -rf /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc.alt.so.12 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc.alt.so.12.9.86 \
+    /usr/local/cuda-12.9/targets/x86_64-linux/lib/libnvrtc-builtins.alt.so.12.9 \
```
leoll2, line 36: Are you sure these files are not redistributable? According to the [documentation](https://docs.nvidia.com/cuda/eula/index.html#attachment-a):

> The following CUDA Toolkit files may be distributed with applications developed by you, **including certain variations of these files that have version number or architecture specific information embedded in the file name**

The table contains `libnvrtc.so` and similar files; based on my interpretation, `libnvrtc.alt.so.12` is allowed since it's a variation of `libnvrtc.so` for CUDA 12. I believe `.alt` doesn't make any difference, it's in fact the same library (same implementation) built in a different way.

image

AlexanderBarabanov: Honestly, I am not 100% sure that we can treat `.alt` as "architecture specific information" - we will double-check it internally.
But if there are no degradations observed in e2e - i would propose to proceed with this removal now as internal process may take some time.

interactive_ai/workflows/train/trainer/gpu/Dockerfile
jcchr: corrected

interactive_ai/workflows/train/trainer/xpu/Dockerfile
leoll2: If Triton redistributes these files, then it means they have a suitable license to do so. I'm not an expert in legal matters, but if triton authors release their package with MIT license and no further disclaimer/notice, I guess you as a client can assume that the whole thing is MIT.

interactive_ai/workflows/train/trainer/gpu/Dockerfile
leoll2: The problem is that e2e don't cover all the possible degradations that could result from this change. An algorithm could become e.g. 10% slower because a linear algebra optimization module is no longer available, and the tests would still pass. Imho we should verify first such this licensing issue is real and there's no other workaround, before merging this PR.

leoll2 (approved):
Approving conditionally to performance regression testing.

PR #1459, opened 2025-11-04
fix(smart-tools): Fix model loading if cache is full

MarkRedeman:
## 📝 Description

Gracefully handle cache.put failures (e.g. low storage) and fall back to direct fetch when the Cache API is unavailable.

## ✨ Type of Change

## 🧪 Testing Scenarios

- [x] ✅ Tested manually

### DIFF
--- a/web_ui/packages/smart-tools/src/utils/tool-utils.ts
+++ b/web_ui/packages/smart-tools/src/utils/tool-utils.ts
@@ -130,7 +130,7 @@ export const loadSource = async (source: string, cacheKey = 'general'): Promise<
 
     // Putting the model in cache might fail if the user does not have enough storage
     try {
-        await cache.put(source, response);
+        await cache.put(source, response.clone());
     } finally {
         return response;
     }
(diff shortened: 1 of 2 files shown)


web_ui/packages/smart-tools/src/utils/tool-utils.ts
```diff
@@ -121,9 +121,18 @@ export const loadSource = async (source: string, cacheKey = 'general'): Promise<
...
+
+    const response = await self.fetch(source);
 
-    if (!(await cache.match(source))) {
-        await cache.put(source, await self.fetch(source));
+    // Putting the model in cache might fail if the user does not have enough storage
+    try {
+        await cache.put(source, response);
+    } catch {
+        return response;
```
dwesolow, line 135: I guess line 138 is no longer needed? We can just return `response`, right?

MarkRedeman: You're right! I initially thought this would be a good extra way to make sure `caches.match` would work. But in hindsight I think it's safer if we do this in a `finally` block. Will update

jpggvilaca (dismissed):

jpggvilaca (approved):

dwesolow (approved):

PR #1460, opened 2025-11-04
chore: Update virtualized horizontal grid

dwesolow:
## 📝 Description

This PR updated virutalized horizontal grid so it's more customizable.

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

- [X] ✅ Tested manually

## ✅ Checklist

- [X] 🔍 PR title is clear and meaningful

### DIFF
--- a/web_ui/packages/ui/src/virtualized-horizontal-grid/virtualized-horizontal-grid.tsx
+++ b/web_ui/packages/ui/src/virtualized-horizontal-grid/virtualized-horizontal-grid.tsx
@@ -1,7 +1,7 @@
 // Copyright (C) 2022-2025 Intel Corporation
 // LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
 
-import { ReactNode } from 'react';
+import { ComponentProps, ReactNode } from 'react';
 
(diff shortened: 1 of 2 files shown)


MarkRedeman (approved):
Approving to unblock, though I'm a bit worried that using this component won't add us much benefits: it might be better to use the react-aria components directly.

The thing I'm afraid of is that we will end up having to update this component every time that we need to expose another prop, ending up with a complex component that mainly forwards its props to the underlying aria components, but with less flexibility.

dwesolow:
> Approving to unblock, though I'm a bit worried that using this component won't add us much benefits: it might be better to use the react-aria components directly.
> 
> The thing I'm afraid of is that we will end up having to update this component every time that we need to expose another prop, ending up with a complex component that mainly forwards its props to the underlying aria components, but with less flexibility.

Totally agree but for now we use that only in two places so was easier to extend it. If we want to use aria components directly we will need to expose `HorizontalLayout` which is our custom one, aria components does not support it.

jpggvilaca (approved):

PR #1531, opened 2025-11-20
Allowing definition of k3s node name

### LINKED ISSUE #1473: K3s node name enforced during installation
As it can be seen in #915 , not setting the node name for K3s causes problems when changing the host name on the environment.
To prevent this situation, we should add the parameter `--node-name geti \` during its (k3s) installation.
We should not add it for Geti upgrades because it will interfere too much with current deployments (duplicate nodes that need to be taken care of).

jcchr:
## 📝 Description

- Fixes #1473

## ✨ Type of Change

- [ ] 🔒 **Security update**
- [ ] 🧪 **Tests**

## 🧪 Testing Scenarios

Check that if the --node-name parameter of the install command is given - k3s gets a given name. Check if everything works like previously - when the parameter is not given. Check if the name is still kept after upgrade of k3s made using installer.

- [ ] ✅ Tested manually

## ✅ Checklist

### DIFF
--- a/platform/services/installer/app/commands/install.py
+++ b/platform/services/installer/app/commands/install.py
@@ -359,7 +359,8 @@ def execute_installation(config: InstallationConfig) -> None:  # noqa: C901, RUF
         logger.info(InstallCmdTexts.k3s_installing)
         try:
             with click_spinner.spinner():
-                install_k3s()
+                node_name = config.node_name.value if config.node_name else None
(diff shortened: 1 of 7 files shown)


piotrgrubicki (changes requested):
To prevent future problems, the node name cannot be optional. Let's hardcode it.

piotrgrubicki (approved):

